PR #194, opened 2026-07-06
fix: signup flow reliability and error messaging

### LINKED ISSUE #182: bug : User Unable to Create Account Due to Server Error .
## Description
Users are currently unable to complete the registration process. Upon clicking the "Create account" button, the system triggers a generic "Server error" message, preventing successful sign-up.

### Impact
- User Acquisition: This is a blocking issue that prevents any new users from signing up for the platform.

- Conversion: Directly halts the user onboarding funnel.

### Expected Behavior
- When a user fills in valid registration details and clicks "Create account," the system should successfully process the request and create the new user profile.

- In the event of an actual server-side failure, the system should log the specific error internally and provide the user with a helpful message, rather than a vague "Server error."

## # Steps to Reproduce
- Navigate to the sign-up page.

- Enter valid user information (Name, Email, Password meeting all complexity requirements).

- Click the "Create account" button.

Image

sukanya-2006:
## Description

Fixes the misleading "Server error" / false success messaging in the sign-up 
flow (Closes #182 ).

Previously:
- When account creation failed, the frontend only showed a generic "Server 
  error" — even though the backend already computed a more specific detail 
  and never sent it through.
- If the OTP email failed to send, the app silently treated it as non-fatal 
  and told the user "OTP sent to your email" regardless — leaving users 
  stuck with no idea anything went wrong.

## Changes

- **`route.ts`** — now checks if the email actually sent; returns an honest 
  partial-success (207) with a clear message if it didn't
- **`sign-up-form.tsx`** — now shows the specific backend message instead of 
  the generic fallback, and correctly handles the 207 case (since it still 
  falls in the "ok" status range)
- **`email.ts`** — the local dev console-log fallback (used when 
  `RESEND_API_KEY` is missing) is now gated to non-production only. In 
  production, a missing key now correctly returns a failure instead of 
  silently claiming success

## Testing

- Verified full signup → OTP → verify → sign-in flow works locally
- Simulated an email failure with an invalid `RESEND_API_KEY`, confirmed the 
  new honest error message shows correctly
- `npm run build` passes with no errors

## For maintainer to check

While testing the **live site**, I signed up with real Gmail addresses 
multiple times and never received an OTP email. Could you confirm 
`RESEND_API_KEY` is actually set in Vercel's environment variables? If it's 
missing, that's likely why real users aren't getting OTPs at all — this PR 
makes that failure honest, but doesn't fix delivery itself since I don't 
have access to your Resend/Vercel config to verify further.

### DIFF
--- a/src/app/api/auth/signup/route.ts
+++ b/src/app/api/auth/signup/route.ts
@@ -66,10 +66,15 @@ export async function POST(req: NextRequest) {
     });
 
     // Send OTP email
-    try {
-      await sendOTPEmail(email, otp);
-    } catch (emailError) {
-      console.error("Signup email error (non-fatal):", emailError);
+    const emailResult = await sendOTPEmail(email, otp);
+
+    if (!emailResult.success) {
+      console.error("Signup email error:", emailResult.error);
+      return NextResponse.json({
+        ok: true,
+        userId: user.id,
+        error: "Account created, but we couldn't send the verification email. Please use 'Resend OTP' or contact support.",
+      }, { status: 207 }); // 207 = Multi-Status: partial success
     }
 
     return NextResponse.json({ ok: true, userId: user.id, message: "OTP sent to your email" });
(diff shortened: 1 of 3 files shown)


singh-odyssey (approved):

singh-odyssey:
@sukanya-2006 
Hey there ,
RESEND_API_KEY is already set in vercel env .
This fix will be pushed in prod .
Then you may test it again ,

Issue #317, opened 2026-07-31
TravelHeatmap imports "react-map-gl/maplibre" (not installed) — typecheck & build fail on main


vedant7007:
## Description

`TravelHeatmap` imports a module that isn't installed, so `tsc --noEmit` fails and a production `next build` would fail webpack module resolution — `main` is currently red on type-check/build.

`src/components/spatial/TravelHeatmap.tsx:4`:
```ts
import Map, { Source, Layer } from "react-map-gl/maplibre";
```

`package.json` declares `@vis.gl/react-maplibre@^8` (the v8 rename of `react-map-gl`). `react-map-gl` itself isn't a dependency and isn't installed, so this import can't resolve.

## Steps to Reproduce

1. `npm install`
2. `npx tsc --noEmit`
3. → `TravelHeatmap.tsx(4,36): error TS2307: Cannot find module 'react-map-gl/maplibre'`

(`next build` in CI hits the same resolution failure.)

## Expected behavior

Import from the installed package: `@vis.gl/react-maplibre` (the v8 API still exports `Map`, `Source`, `Layer`). One-line change, after which the Travel Heatmap feature actually builds/renders.

## Environment

`main` (latest). CI (`.github/workflows/ci.yml`) runs lint + `tsc --noEmit` + `next build`, all of which trip on this.

## Additional context

The feature landed in #296 / PR #303; PR #315 ("Fix lint build") didn't correct this import, and it's still on HEAD. No open issue reports it.

Happy to send the one-line fix if you'd like to assign it to me.

_Contributing as part of Elite Coders Summer of Code (ECSoC 2026)._

PR #330, opened 2026-08-02
fix(matches): remove 100-row cap so total is accurate and all matches reachable (#320)

### LINKED ISSUE #320: C:/Program Files/Git/api/matches caps total at 100 and over-fetches on every page
## Description

The `/api/matches` endpoint caps its result set at 100 rows and paginates in memory over that cap, so once a destination/date window has more than 100 matching tickets, `total`/`hasMore` are wrong and any match past #100 is unreachable. It also over-fetches on every request.

`src/app/api/matches/route.ts:342-411`:
- the query uses `take: 100`
- `const total = scoredMatches.length` → always ≤ 100
- `paginatedMatches = scoredMatches.slice(skip, skip + limit)`
- `hasMore = skip + limit < total`

The full (up to) 100-row set is scored in memory, cached, and re-sliced on every page — even for `page=1&limit=10`.

## Steps to Reproduce

1. Have >100 verified tickets matching a given destination/date window.
2. Call `/api/matches` and page through.
3. `total` maxes out at 100, `hasMore` goes false early, and matches beyond #100 never appear.

## Expected behavior

Push pagination into Prisma (`skip`/`take` plus a real `count()` for `total`), or use the repo's existing cursor helper in `src/lib/pagination.ts` (already used by tickets/routes/admin), so counts are accurate and every match is reachable without loading 100 rows per request.

## Environment

`main` (latest), reproduced from source.

## Additional context

#294 addressed N+1 in match discovery, and #287/PR #299 added cursor pagination to **routes and tickets only** — the matches endpoint's offset-cap bug is separate and still present.

Happy to fix it (real DB-level pagination via the existing helper) if you'd like to assign it to me.

vedant7007:
## Description

`/api/matches` queried with `take: 100`, then scored, sorted and paginated the result set **in memory**. Once a destination/date window had more than 100 matching tickets:
- `total` maxed out at 100 (`const total = scoredMatches.length`),
- `hasMore` went false early,
- and any match past #100 was unreachable.

Relevance is scored in memory (`calculateRelevance` over the current user + target date), so DB-side `skip/take` can't be used without losing the global ranking. The fix is to drop the arbitrary `take: 100` so the full candidate set — already bounded by the `destination` + date-window + status filters in `whereClause` — is ranked and counted correctly.

The scored list is written to the match cache, and the cache-hit branch already slices the full list per page, so the scan runs **once per window** and subsequent pages are served from the cache rather than re-querying. `total` / `hasMore` are now accurate and every match is reachable.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Checklist
- [x] Lints pass (`npm run lint` — clean on the changed file)
- [x] Typechecks (`tsc --noEmit` — no new errors from this change)
- [x] This is already assigned Issue to me, not an unassigned issue.

> CI note: `main` is currently red on `tsc`/`build` due to the unrelated `TravelHeatmap.tsx` stale import (#317, fixed in my PR #321). This change itself is green.

### DIFF
--- a/src/app/api/matches/route.ts
+++ b/src/app/api/matches/route.ts
@@ -339,9 +339,13 @@ export async function GET(req: NextRequest) {
       whereClause.user = userWhere;
     }
 
+    // Relevance is scored/sorted in memory, so the full candidate set (bounded
+    // by the destination + date-window + status filters in whereClause) has to
+    // be fetched to rank correctly and report an accurate total. The result is
+    // cached below, so this scan runs once per window and every subsequent page
+    // is served from the cache rather than re-querying.
     const matches = await prisma.ticket.findMany({
       where: whereClause,
-      take: 100,
       select: {
         id: true,
         destination: true,


vedant7007:
No worries about closing this — I realize removing the `take: 100` cap outright makes the query unbounded, which is a fair thing to be wary of on a hot path.

If you'd still like #320 addressed, I'm happy to reopen with a bounded approach instead: keep a generous hard cap but return an accurate `total` via a separate `count()` (so `hasMore` is correct), or move to true DB-side pagination using the existing `src/lib/pagination.ts` cursor helper that's already used by tickets/routes. Just let me know which direction you'd prefer and I'll send a fresh PR — otherwise no problem at all.

Issue #348, opened 2026-08-08
[SECURITY] Blocked users can still send and receive direct messages


MOHITKOURAV01:
## Summary

Blocking a user does not stop that user from messaging you. `Block` is checked in `/api/matches` and in `POST /api/connections`, but the messaging layer ignores it completely, so an already-open conversation stays fully usable after a block.

## Where

- `src/app/api/messages/route.ts` — `POST` only checks conversation membership (lines 117-128) and never looks at `Block`.
- `src/app/api/messages/route.ts` — `GET` likewise returns the full history regardless of a block.
- `src/app/api/conversations/route.ts` — `GET` lists every conversation the user belongs to, so a blocked person keeps appearing in the sidebar.

## Steps to reproduce

1. User A and User B connect, which creates a conversation.
2. User A blocks User B via `POST /api/user/block`.
3. User B posts to `/api/messages` with that `conversationId`.
4. The message is created, `triggerPusher` fires `private-chat-<id>` and `private-user-<A id>`, and it lands in A's UI in real time.

## Expected

- Sending a message must fail with `403` when either side has blocked the other (blocks are two-way, same as the rule already used in `POST /api/connections`).
- Reading a blocked conversation should not be possible.
- Blocked conversations should be hidden from the conversation list.

## Notes

`POST /api/connections` already implements the two-way rule with an `OR` of `{ blockerId, blockedId }` pairs. That logic is duplicated inline; it is worth extracting into a small shared helper so messaging, connections and any future feature use exactly one implementation.

## Impact

This is the core safety feature of the platform, and it currently does not work for the one surface that matters most. Harassment via DMs continues after a block.

Issue #318, opened 2026-07-31
Test suite is red (3 failing) and CI never runs vitest


vedant7007:
## Description

`vitest` is red on `main` (3 failing tests), and because CI never runs the test suite, nothing catches it. The failures are stale tests left behind by later refactors, not fresh product bugs — but they mean the suite no longer guards the routes it's supposed to.

**`src/app/api/tickets/__tests__/tickets.test.ts:117-146`** — the "creates ticket" test's prisma mock (lines 6-15) only stubs `user.findFirst` and `ticket.create`, but the route now calls `prisma.ticket.findMany` in `findDuplicateTicket` (`src/app/api/tickets/route.ts:98`). That's undefined in the mock → throws → 500. The test also still expects `200`, but the route now returns `201` (`route.ts:253`). So it fails with `expected 500 to be 200`. This is drift from the idempotency/dedup refactor (PRs #288/#300).

**`src/components/__tests__/sign-in-form.test.tsx:52,71`** — both tests time out at 5000ms (async issue).

**CI gap** — `.github/workflows/ci.yml:32-38` runs lint / `tsc` / build only; there's no `npm test` step, so a red suite never blocks a PR.

## Steps to Reproduce

1. `npm install`
2. `npx vitest run`
3. → 3 failing / 146 passing (2 files failed).

## Expected behavior

- Update `tickets.test.ts` to mock `prisma.ticket.findMany` (+ the idempotency/cache lookups) and expect `201`, so it guards the current route again.
- Fix the `sign-in-form` async timeout.
- Add a `Test` step (`npm test`) to `ci.yml` so the suite can't silently rot.

## Environment

`main` (latest), local `vitest run`.

## Additional context

Prior test PRs #230/#231 fixed different suites; no open issue tracks these current failures or the missing CI test step. Happy to take this end-to-end if you'd like to assign it to me.

_Contributing as part of Elite Coders Summer of Code (ECSoC 2026)._

PR #355, opened 2026-08-08
perf(messages): paginate conversation history

### LINKED ISSUE #350: [PERF] GET /api/messages returns the entire conversation history with no pagination
## Summary

`GET /api/messages?conversationId=...` runs an unbounded `findMany` and returns every message in the conversation, each one joined with its sender and its shared route (including `encodedPolyline`, which is a `@db.Text` column).

## Where

`src/app/api/messages/route.ts` lines 35-70:

```ts
const messages = await prisma.message.findMany({
  where: { conversationId },
  orderBy: { createdAt: "asc" },
  include: { sender: {...}, route: {...} },
});
```

There is no `take`, no `skip`, and no cursor. `src/components/ChatInterface.tsx` calls this on every conversation switch and renders the whole array.

## Why this matters

- Payload grows without bound. A conversation with a few thousand messages, some of them shared routes with long encoded polylines, is a multi-megabyte JSON response on every conversation click.
- The response is re-fetched from scratch each time the user switches conversations.
- It is a trivially cheap way for one user to generate very large responses from the server.

## Expected

Use the cursor pagination helpers that already exist in `src/lib/pagination.ts` (`parsePaginationParams`, `buildTimestampCursorWhere`, `createPaginatedResponse`) — the same pattern `/api/routes` and `/api/tickets` use. Newest messages first, `limit` capped by `MAX_PAGE_LIMIT`, `nextCursor` for loading older messages.

The existing `{ messages: [...] }` response key should be kept alongside the new `items`/`pagination` fields so the current client keeps working, which is what `/api/routes` does with its `routes` alias.

The chat UI needs a matching "Load older messages" control, and it must keep rendering oldest-to-newest even though the API returns newest-first.

## Related

`GET /api/conversations` has the same unbounded-`findMany` shape, but that list grows much more slowly; message history is the urgent one.

MOHITKOURAV01:
Closes #350

## Problem

`GET /api/messages` had no `take`, no `skip` and no cursor:

```ts
const messages = await prisma.message.findMany({
  where: { conversationId },
  orderBy: { createdAt: "asc" },
  include: { sender: {...}, route: {...} },
});
```

Every message in the thread came back on every conversation switch, each one joined with its sender and its shared route — including `encodedPolyline`, a `@db.Text` column. A long thread with a few shared routes is a multi-megabyte response, re-fetched each time the user clicks a different conversation.

## API

Paged with the helpers already in `src/lib/pagination.ts`, the same way `/api/routes` and `/api/tickets` do it. Newest first, `limit` defaulting to 20 and capped at `MAX_PAGE_LIMIT`, cursor on `(createdAt, id)`.

```json
{
  "items":      [ /* newest first — cursor lines up with the last element */ ],
  "pagination": { "limit": 20, "nextCursor": "…", "hasMore": true },
  "messages":   [ /* same page, oldest first — render order */ ]
}
```

Both orderings are returned on purpose. `items`/`pagination` are the pagination contract; `messages` is the alias the current client already reads, re-sorted oldest-first because that is the order a transcript renders in. Same "keep the old key working" approach `/api/routes` took with its `routes` alias, so nothing that reads `data.messages` breaks.

A malformed `limit` or `cursor` now returns `400` via `PaginationError` instead of falling into the generic `500`.

The `sender`/`route` include was duplicated between `GET` and `POST`. It is hoisted into one `MESSAGE_INCLUDE` constant — the chat renders both through the same component, so the two shapes drifting apart would be a real bug.

## Client

`ChatInterface` keeps the cursor from the first page and shows a **Load older messages** control above the transcript while one exists.

- Each page is prepended, filtered against ids already present, because a message can arrive over Pusher while the request is in flight.
- The scroll-to-newest effect is suppressed for exactly one pass when history is prepended (`skipScrollToBottom` ref), otherwise loading older messages would yank the viewport back to the bottom.
- The control disappears once `nextCursor` is `null`.

## Docs

`docs/API.md` had no messaging section at all. Added `GET /api/messages` (with the parameter table and both orderings explained), `POST /api/messages` and `GET /api/conversations`.

## Tests

`npx vitest run src/app/api/messages` → 13 passed.

`src/app/api/messages/__tests__/pagination.test.ts` is new (10 cases): `take: limit + 1` and the `desc, desc` ordering, the bounded default, the `MAX_PAGE_LIMIT` cap, both orderings in the response, `hasMore` + `nextCursor` on a full page, no cursor on the last page, the cursor `where` clause, `400` for a bad cursor and a bad limit, and that the membership `404` still short-circuits before any message query.

`npm run lint` is clean, no new `tsc` errors.

**On the client change:** I tried to add a `@testing-library/react` test for the load-older flow and could not get `ChatInterface` to mount under the current jsdom setup — it fails with an "element type is invalid" from a child that I could not pin down, and mocking `TripBoard`, `RoutePickerModal`, `RouteShareCard` and `next/image` did not clear it. Rather than ship a test I had to contort, I left the component untested here and verified the flow by hand. Happy to follow up separately if somebody knows what that component needs to render in tests.

## Notes for review

- No schema change. `Message` already has `@@index([conversationId])`; the sort is `(createdAt desc, id desc)`, so a composite `@@index([conversationId, createdAt, id])` would help on very large threads — left out of this PR because it needs a migration and belongs in its own change.
- This branch only touches the `GET` body and the include constant in `/api/messages`, so it does not collide with the block checks in #348.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -241,6 +241,69 @@ Response:
 }
 ```
 
+## Messaging endpoints
+
+### `GET /api/messages?conversationId=<id>`
+
+Return a page of the conversation transcript. The caller must be a participant,
+otherwise the endpoint responds `404`.
+
+Query parameters:
+
+| Name | Default | Notes |
+| --- | --- | --- |
+| `conversationId` | — | Required. |
+| `limit` | `20` | Capped at `100`. |
+| `cursor` | — | Opaque cursor from a previous `pagination.nextCursor`. |
+
+Response:
+
+```json
+{
+  "items": [{ "id": "msg-3", "text": "See you at the gate", "createdAt": "..." }],
+  "pagination": {
+    "limit": 20,
+    "nextCursor": "eyJ2ZXJzaW9uIjoxLCJ0aW1lc3RhbXAiOiIuLi4iLCJpZCI6Im1zZy0yIn0",
+    "hasMore": true
+  },
+  "messages": [{ "id": "msg-1", "text": "Landing at 6", "createdAt": "..." }]
+}
+```
+
+Two orderings are returned deliberately:
+
+- `items` is newest-first, matching the query order, so `pagination.nextCursor`
+  lines up with the last element.
+- `messages` is the same page re-sorted oldest-first, which is the order a chat
+  transcript is rendered in. Pass `pagination.nextCursor` back as `cursor` to
+  walk further into the past and prepend the result.
+
+`400` is returned for a malformed `limit` or `cursor`.
+
+### `POST /api/messages`
+
+Send a message to a conversation the caller belongs to. Either `text` or
+`routeId` must be present.
+
+Request body:
+
+```json
+{
+  "conversationId": "conv-1",
+  "text": "Booked the 7am bus",
+  "routeId": "route-1"
+}
+```
+
+A `routeId` must belong to the sender; otherwise the endpoint responds `403`.
+On success the created message is broadcast over Pusher to the conversation
+channel and to each participant's personal channel.
+
+### `GET /api/conversations`
+
+List the caller's conversations, most recently updated first, each with the
+other participant and the latest message for the sidebar.
+
 ## AI chat endpoint
 
 ### `POST /api/chat`
--- a/src/app/api/connections/__tests__/route.test.ts
+++ b/src/app/api/connections/__tests__/route.test.ts
@@ -16,6 +16,10 @@ vi.mock("@/lib/prisma", () => ({
       findFirst: vi.fn(),
       create: vi.fn(),
     },
+    block: {
+      findFirst: vi.fn(),
+      findMany: vi.fn(),
+    },
   },
 }));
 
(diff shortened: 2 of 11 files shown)


MOHITKOURAV01:
Heads up on merge order for the three messaging PRs.

#353, this one and #357 all rewrite parts of `src/app/api/messages/route.ts`. All three merge cleanly into `main` as they are, but whichever lands first would leave the other two conflicting, so I have stacked them in the order they should land:

```
main → #353 (block enforcement) → #355 (this one) → #357 (rate limiting)
```

This branch is rebased on top of #353, so its commit shows here for now and GitHub will drop it from the diff automatically once #353 is merged. Reviewing the `perf(messages): paginate conversation history` commit on its own gives the change this PR is actually about.

Verified locally: merging all five open branches into `main` in the order #353 → #355 → #357 → #354 → #356 produces no conflicts, and the combined tree passes `npm run lint`, `npx tsc --noEmit` and the test suite with the same pre-existing failures as `main` (admin-ticket `PATCH` exports from #318, the `hashOTP` mock in the signup test, and the tickets test).

singh-odyssey (approved):

Issue #349, opened 2026-08-08
[BACKEND] /api/user/block is write-only: no unblock endpoint and no way to list blocked users


MOHITKOURAV01:
## Summary

`/api/user/block` only exposes `POST`. There is no way to undo a block and no way to find out who you have blocked, so the block list is write-only and permanent from the API's point of view.

## Where

`src/app/api/user/block/route.ts` — the file exports `POST` and nothing else.

## Missing behaviour

1. **`DELETE /api/user/block`** — remove an existing `Block` row for `(blockerId = me, blockedId = target)`. Should return `404` when no such block exists and must never let a user delete somebody else's block.
2. **`GET /api/user/block`** — return the list of users the caller has blocked (id, name, image, blockedAt) so a "Blocked users" settings screen can be built. This should be paginated with the existing cursor helpers in `src/lib/pagination.ts`, the same way `/api/routes` and `/api/tickets` already do it.

## Smaller problems in the current `POST`

- `prisma.user.findUnique({ where: { id: blockedId } })` selects every column of the target user (including `passwordHash`, `otp`, `resetToken`) just to check existence. It should `select: { id: true }`.
- A soft-deleted user (`isDeleted: true`) can still be blocked.
- Blocking somebody does not clean up a pending `ConnectionRequest` between the two users, so the request stays in their "incoming" list.

## Impact

Users who block somebody by mistake are stuck. There is also no UI possible for reviewing blocks, which is a common requirement for a safety feature.

PR #357, opened 2026-08-08
feat(security): rate limit the user-content write endpoints

### LINKED ISSUE #352: [SECURITY] Rate limiting is only applied to auth routes; messaging, connections, reports and blocks are unthrottled
## Summary

`src/lib/rate-limit.ts` is a solid Redis-backed limiter with tests, but it is only wired into six routes, all of them auth-related plus `/api/chat`:

```
src/app/api/auth/signup/route.ts
src/app/api/auth/verify-otp/route.ts
src/app/api/auth/forgot-password/route.ts
src/app/api/auth/resend-otp/route.ts
src/app/api/auth/[...nextauth]/route.ts
src/app/api/chat/route.ts
```

Every endpoint that writes user-generated content is unthrottled.

## Unthrottled write endpoints

| Endpoint | What an authenticated caller can do in a loop |
| --- | --- |
| `POST /api/messages` | Create unlimited `Message` rows and fire one Pusher event per message per participant. Pusher message quota is billable. |
| `POST /api/connections` | Spam connection requests; each `send` creates a `Notification` row and a Pusher event on the receiver's channel. |
| `POST /api/user/report` | Create unlimited `Report` rows against the same user, which is itself an abuse vector against the moderation queue. |
| `POST /api/user/block` | Unbounded writes. |

`POST /api/tickets` is partly protected by the idempotency key and the Cloudinary upload cost, but it has no request-rate ceiling either.

## Also: the rate-limit constants are copy-pasted

Each of the six protected routes hardcodes its own `namespace`/`limit`/`windowSeconds` literals at the call site. There is no single place to see or tune the policy, and nothing stops two routes from accidentally sharing a namespace.

## Expected

1. A single `src/lib/rate-limit-rules.ts` module that declares the named rules (namespace, limit, window) for every throttled route, with a small `enforceRateLimit(request, rule, subject)` wrapper over `checkRateLimit` so call sites stay one line.
2. Apply it to `POST /api/messages`, `POST /api/connections`, `POST /api/user/report` and `POST /api/user/block`, keyed on the session user id plus IP via the existing `getRateLimitIdentifier`.
3. Keep the current fail-open behaviour when Redis is unavailable — `checkRateLimit` already returns `bypassed: true` in that case and that is the right call for a social app.
4. `429` responses must carry `Retry-After` and the `X-RateLimit-*` headers, which `rateLimitExceededResponse` already does.

Unit tests should cover both the allowed and the exceeded path for at least one of the newly protected routes.

MOHITKOURAV01:
Closes #352

## Problem

`src/lib/rate-limit.ts` is a solid Redis-backed limiter with its own tests, but it was wired into six routes — five auth ones and `/api/chat`. Every endpoint that writes user-generated content was unthrottled:

| Endpoint | What a loop costs |
| --- | --- |
| `POST /api/messages` | one `Message` row **and** one Pusher event per participant, per iteration |
| `POST /api/connections` | a `Notification` row plus a Pusher event on somebody else's channel |
| `POST /api/user/report` | unbounded rows in the moderation queue |

Pusher quota is billable, so the messaging one is a direct cost-abuse path.

Separately, each protected route hardcoded its own `namespace`/`limit`/`windowSeconds` at the call site. Nothing showed the overall policy in one place and nothing stopped two routes from sharing a namespace — sharing one means sharing a counter, so traffic to one would lock the other out.

## `src/lib/rate-limit-rules.ts` (new)

One table of named policies plus a thin wrapper:

```ts
const rateLimit = await enforceRateLimit(req, "messageSend", userId);
if (!rateLimit.allowed) return rateLimitExceededResponse(rateLimit);
```

`satisfies Record<string, RateLimitRule>` keeps the entries type-checked while preserving literal key types, so `ruleName` is a union and a typo is a compile error rather than a silently-new namespace.

## Policies applied

| Rule | Limit | Keyed on |
| --- | --- | --- |
| `messageSend` | 30 / min | session user id + IP |
| `connectionAction` | 20 / min | session user id + IP |
| `userReport` | 5 / 10 min | session user id + IP |

30/min is comfortable for a fast typist in a live conversation and still cuts a scripted flood dead. Reports are deliberately the tightest — filing them in bulk is itself an abuse vector, and there is a test asserting `userReport.limit < messageSend.limit` so a later edit cannot quietly invert that.

The check runs **before** the body is read on `/api/messages` and `/api/connections`, so a throttled caller cannot make the server parse a payload.

## Existing routes moved onto the table

The five auth routes and `/api/chat` now read from the same table. **No limit changes** — a test pins all six values explicitly, so the move cannot alter policy by accident.

## Fail-open

Unchanged and deliberate: `checkRateLimit` already returns `bypassed: true` when Redis is missing or unreachable, and `enforceRateLimit` inherits it. Locking every user out of messaging because the cache is down is the worse failure for a social app.

## Tests

`npx vitest run src/lib/__tests__/rate-limit-rules.test.ts src/app/api/messages src/app/api/connections src/app/api/user` → 31 passed, 1 pre-existing failure (see below).

- `src/lib/__tests__/rate-limit-rules.test.ts` (new, 8 cases): namespaces are unique, every rule has a positive limit and window, the six migrated policies match their previous values exactly, reports stay tighter than messaging, the rule reaches `checkRateLimit` intact, IP-only fallback without a subject, per-subject counter scoping.
- `src/app/api/messages/__tests__/rate-limit.test.ts` (new, 4 cases): throttled on the sender id rather than the IP alone, `429` with `Retry-After: 17` and `X-RateLimit-Remaining: 0` writing nothing and firing no Pusher event, the body not being read once exceeded, and the headers present on a successful send.
- The existing messaging and connections tests needed a `headers` bag on their hand-rolled mock requests — `getClientIp` reads `x-forwarded-for`. Extracted into a `postRequest` helper in each file; no assertions changed.

`npm run lint` is clean, no new `tsc` errors. Repo-wide the suite has the same 13 failures as `main` (admin-ticket `PATCH` exports, the `hashOTP` mock in the signup test, and the `block` model missing from the connections mock) — all pre-existing and none touched here.

## Notes for review

- `POST /api/user/block` is **not** included. #349 rewrites that file end to end, and adding a limit to it here would create a conflict for no benefit. It should get `enforceRateLimit` in a one-line follow-up once one of the two lands.
- `POST /api/tickets` is also left out: it is already gated by the idempotency key and the Cloudinary upload cost, and adding a request-rate ceiling there deserves its own discussion about what the right number is.


MOHITKOURAV01:
Heads up on merge order for the three messaging PRs.

#353, #355 and this one all touch `src/app/api/messages/route.ts` (and #353 and this one both touch `src/app/api/connections/route.ts`). All three merge cleanly into `main` today, but whichever lands first would leave the others conflicting, so I have stacked them in the intended landing order:

```
main → #353 (block enforcement) → #355 (message pagination) → #357 (this one)
```

Those two commits show in this PR for now; GitHub drops them from the diff as each predecessor merges. The `feat(security): rate limit the user-content write endpoints` commit is the one this PR is about.

Resolving the stack also cleared up a test-mock detail worth flagging: the hand-rolled request objects in the messaging and connections tests had no `headers` bag, which `getClientIp` needs. They now go through a `postRequest` helper in each file — no assertions changed.

Verified locally: merging all five open branches into `main` in the order #353 → #355 → #357 → #354 → #356 produces no conflicts, and the combined tree passes `npm run lint`, `npx tsc --noEmit` and the test suite with the same pre-existing failures as `main`.

singh-odyssey (approved):

Issue #366, opened 2026-08-09
Bug: CI Pipeline Failing on Base Branch (Lint, Build, Test)


singh-odyssey:
## 🚨 The Problem
The Continuous Integration (CI) pipeline (Lint, Build, and Test steps) is currently failing on new Pull Requests. The failures appear to be caused by pre-existing code or configurations in the repository rather than the changes introduced in the new PRs. 

Because the base branch is currently in a broken state, it is blocking new work from being safely merged.

## 👣 Steps to Reproduce
1. Create a new branch from `main` (or `master`).
2. Push a benign change or an empty commit (`git commit --allow-empty -m "Trigger CI"`).
3. Open a Draft PR.
4. Observe the CI pipeline failures (Lint, Build, and/or Test).

## 🎯 Impact
- Blocks all current and incoming Pull Requests.
- Developers cannot verify if their new code breaks the build, as the baseline is already broken.

## ✅ Acceptance Criteria
- [ ] Identify the root cause of the linting, build, and test failures.
- [ ] Fix the underlying code or configuration issues causing the failures.
- [ ] Ensure the CI pipeline successfully passes on the base branch.
- [ ] (Optional) Add a safeguard to prevent merging PRs if the base branch CI is failing in the future.

vedant7007:
Root cause of the base-branch build failure: four auth routes have a duplicate `const rateLimit` declaration plus calls to `checkRateLimit`/`getRateLimitIdentifier` that aren't imported in those files.

- `src/app/api/auth/signup/route.ts` (L30-31)
- `src/app/api/auth/resend-otp/route.ts`
- `src/app/api/auth/verify-otp/route.ts`
- `src/app/api/auth/forgot-password/route.ts`

Each has:
```ts
const rateLimit = await enforceRateLimit(_req, ...);   // correct helper
const rateLimit = await checkRateLimit({ ... });        // redeclare + undefined fn
```
`Cannot redeclare block-scoped variable 'rateLimit'` + `checkRateLimit is not defined` fail `tsc`/build, and this also breaks signup/OTP-verify/OTP-resend/password-reset at runtime. Fix: delete the second `checkRateLimit(...)` block in each of the four files, keeping only the `enforceRateLimit(...)` line. Happy to open the PR if useful.

_Contributing as part of Elite Coders Summer of Code (ECSoC 2026)._

PR #376, opened 2026-08-10
fix: GET /api/matches sends a self-notification on every filter change

vedant7007:
Closes #369

### Problem
`GET /api/matches` created a `MATCH_FOUND` notification whenever `scoredMatches.length > 0`. Because each distinct filter combination is a separate cache key (a fresh cache miss), a user adjusting filters received a "New traveller match found" notification for every one of their own searches.

### Fix
Removed the `createNotification` call from the search path (and its now-unused import). Match notifications should be driven by ticket creation/verification, not by the searchers own GET.

### Testing
- `npx tsc --noEmit` clean on the changed route.

_Contributing as part of Elite Coders Summer of Code (ECSoC 2026)._

Issue #368, opened 2026-08-09
[Bug] Admin ticket PATCH handler missing — [id]/route.ts is a copy of the history route (tickets can never be verified)


vedant7007:
## Description

`src/app/api/admin/tickets/[id]/route.ts` is byte-identical to `src/app/api/admin/tickets/[id]/history/route.ts` — it exports only a `GET` (audit-log reader) and has **no `PATCH`**. So an admin has no way to verify/reject a ticket, and since `GET /api/matches` only matches `status: "VERIFIED"` tickets, the entire matching workflow stays permanently empty.

```
$ grep "export async function" src/app/api/admin/tickets/[id]/route.ts
export async function GET(   # only GET — no PATCH
```

The test suite expects a `PATCH` here:
- `__tests__/route-exports.test.ts` asserts `typeof ticketRoute.PATCH === "function"`
- `__tests__/cache-invalidation.test.ts` imports `{ PATCH }` and calls it with `{ status: "VERIFIED" | "REJECTED" }`, expecting `invalidateMatchCachesForTicket(...)` and `createNotification(...)`.

Both tests currently fail (`PATCH` is undefined).

## Steps to Reproduce
1. As an admin, `PATCH /api/admin/tickets/<id>` with `{ "status": "VERIFIED" }`.
2. → `405 Method Not Allowed`. Tickets can never leave `PENDING`, so `/api/matches` never returns anyone.

## Root cause
A merge/copy overwrote the real `PATCH` handler in `[id]/route.ts` with a duplicate of the history route.

## Fix
Restore the `PATCH` handler: admin check → load ticket → `prisma.ticket.update({ status })` → `invalidateMatchCachesForTicket(...)` → `createNotification(...)`. `__tests__/cache-invalidation.test.ts` documents the exact expected behavior/signature.

Happy to fix if assigned.

_Contributing as part of Elite Coders Summer of Code (ECSoC 2026)._

PR #385, opened 2026-08-10
fix(users): stop exposing email addresses from user search

### LINKED ISSUE #380: [Security] GET /api/users leaks every user's email address and allows account enumeration via ?search=
### Summary

`GET /api/users` returns every matched user's **email address** to any authenticated caller, and the `?search=` filter matches against the `email` column. Together that turns the user-search endpoint into an address-book dump and an email-verification oracle.

### Where

`src/app/api/users/route.ts`

```ts
if (search) {
  whereClause.OR = [
    { name:     { contains: search, mode: "insensitive" } },
    { email:    { contains: search, mode: "insensitive" } },   // <- enumeration oracle
    { location: { contains: search, mode: "insensitive" } },
  ];
}

const users = await prisma.user.findMany({
  where: whereClause,
  select: {
    id: true,
    name: true,
    email: true,        // <- PII returned to every caller
    image: true,
    bio: true,
    location: true,
    createdAt: true,
  },
  ...
});
```

### Why this is a problem

1. **PII disclosure.** Nothing else in the product exposes another user's email. `/api/conversations`, `/api/connections` and `/api/matches` all select `id/name/image/bio/location` only. This endpoint is the odd one out, and a single `GET /api/users?limit=100` walked with the cursor dumps the whole user table's emails.
2. **Account enumeration.** `email contains <term>` lets anyone confirm whether a given address has an account, and worse, harvest by domain — `?search=@gmail.com`, `?search=@company.com`.
3. **Blocked users are not filtered.** `src/lib/blocking.ts` exists precisely for this and `/api/conversations` already calls `getBlockedUserIds()`. User search does not, so somebody you blocked still shows up in your search results and you still show up in theirs.
4. `whereClause` is typed `any`, which is what let the `email` predicate slip in unnoticed.

### Steps to reproduce

1. Sign in as any user.
2. `curl -b <session cookie> 'http://localhost:3000/api/users?limit=100'`
3. Every returned object contains an `email` field for a user you have no relationship with.
4. `curl -b <session cookie> 'http://localhost:3000/api/users?search=gmail.com'` returns everybody with a Gmail address.

### Expected

- `email` is never included in the response payload.
- Search matches on `name` and `location` only.
- Users on either side of a `Block` are excluded from results.
- The `where` clause is typed, not `any`.

### Notes

`src/app/api/users/__tests__/route.test.ts` currently asserts that `email` *is* selected and *is* searched, so those assertions need to be updated along with the fix.

MOHITKOURAV01:
Closes #380

## The problem

`GET /api/users` selected `email` into every result and matched the `?search=` term against the email column:

```ts
whereClause.OR = [
  { name:     { contains: search, mode: "insensitive" } },
  { email:    { contains: search, mode: "insensitive" } },
  { location: { contains: search, mode: "insensitive" } },
];

const users = await prisma.user.findMany({
  where: whereClause,
  select: { id: true, name: true, email: true, ... },
});
```

Two separate problems in one query:

1. Any authenticated caller could walk the cursor with `limit=100` and dump the whole user table's email addresses. No screen in the product shows another user's email — `/api/conversations`, `/api/connections` and `/api/matches` all select `id/name/image/bio/location`.
2. `email contains <term>` is an enumeration oracle. `?search=@gmail.com` harvests by domain; `?search=someone@example.com` answers whether that address has an account.

## What this changes

**The leak.** `email` is out of the projection and out of the search predicate. Search matches `name` and `location` — what you'd actually use to find a travel companion.

**Blocked users.** Search never consulted `src/lib/blocking.ts`. Blocking is symmetric, and `/api/conversations` already filters with `getBlockedUserIds()`; user search was the remaining place a blocked user could still surface. Same helper, one query, applied to the exclusion list.

**A pagination bug found while writing the tests.** The cursor filter and the search filter were both spread into the `where` object as `OR`, so the search clause overwrote the cursor clause:

```ts
const whereClause: any = {
  ...(cursorWhere ?? {}),   // { OR: [...cursor...] }
};
if (search) {
  whereClause.OR = [...];  // silently replaces it
}
```

`?search=goa&cursor=<next>` therefore returned page one forever. Both groups now sit under `AND`.

**Typing.** `whereClause` was `any`, which is what let the `email` predicate through unnoticed. It's `Prisma.UserWhereInput` now.

**Input handling.** `search` is trimmed, a whitespace-only term is ignored rather than running a `contains ""` scan, and terms over 100 characters get a 400. `hasMore` is returned alongside `nextCursor`.

## Tests

The existing file asserted the old behaviour — that `email` *was* selected and *was* searched — so those assertions are inverted. Its `buildTimestampCursorWhere` mock also took the cursor as its first argument while the real signature is `(field, cursor)`, so it was handed the string `"createdAt"`, treated it as a cursor and returned a filter with `Date { NaN }` in it on every call. That's why 3 of the 8 tests in this file fail on `main`. Fixed, so the cursor assertions mean something.

8 tests → 19. Everything the old file covered is still covered.

## Verification

- `npx vitest run src/app/api/users` — 19 passed
- Full suite goes from 9 failing files / 23 failing tests on `main` to 8 / 20 — this branch fixes 3 pre-existing failures and introduces none
- `npx tsc --noEmit` reports nothing new for this path (the remaining errors are the base-branch breakage tracked in #366 / #379)

## Note on the base branch

`main` does not currently typecheck — five auth routes have a duplicate `rateLimit` declaration (#366, fixed by #379). Nothing in this PR touches those files, so it should rebase cleanly once #379 lands.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -123,6 +123,53 @@ Request body:
 }
 ```
 
+### `GET /api/users`
+
+Search other travellers. Requires an authenticated session.
+
+Query parameters:
+
+| Parameter | Type    | Default | Notes |
+| --------- | ------- | ------- | ----- |
+| `search`  | string  | —       | Trimmed. Matched case-insensitively against `name` and `location`. Max 100 characters. |
+| `limit`   | integer | `20`    | Capped at 100. |
+| `cursor`  | string  | —       | Opaque cursor taken from a previous response's `nextCursor`. |
+
+Response:
+
+```json
+{
+  "users": [
+    {
+      "id": "clx...",
+      "name": "Asha Sharma",
+      "image": null,
+      "bio": "Travel enthusiast",
+      "location": "Delhi",
+      "createdAt": "2026-01-04T10:12:00.000Z"
+    }
+  ],
+  "nextCursor": "eyJ2ZXJzaW9uIjox...",
+  "hasMore": true
+}
+```
+
+Privacy rules this endpoint enforces:
+
+- **Email addresses are never returned and are never searched.** Matching on
+  `email` would turn the endpoint into an account-enumeration oracle, and
+  returning it would expose an address that no screen in the product shows.
+- Soft-deleted accounts (`isDeleted: true`) are excluded.
+- The caller is excluded from their own results.
+- Users on either side of a `Block` are excluded, matching the behaviour of
+  `GET /api/conversations`.
+
+Typical responses:
+
+- `200 OK` with the page of results
+- `400 Bad Request` for an invalid `limit`/`cursor` or an over-long `search`
+- `401 Unauthorized` without a session
+
 ## Ticket endpoints
 
 ### `POST /api/tickets`
--- a/src/app/api/users/__tests__/route.test.ts
+++ b/src/app/api/users/__tests__/route.test.ts
@@ -7,15 +7,22 @@ vi.mock("@/lib/auth", () => ({
   })),
 }));
 
+vi.mock("@/lib/blocking", () => ({
+  getBlockedUserIds: vi.fn(async () => []),
+}));
+
 vi.mock("@/lib/pagination", () => ({
   parsePaginationParams: vi.fn((searchParams) => ({
     limit: Math.min(Number.parseInt(searchParams.get("limit") || "20", 10), 100),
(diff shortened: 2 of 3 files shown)


MOHITKOURAV01:

singh-odyssey (approved):

PR #386, opened 2026-08-10
fix(notifications): paginate the list and give it a way to be cleared

### LINKED ISSUE #381: [Bug] GET /api/notifications returns every notification ever created — unpaginated, includes expired rows, and cannot be cleared
### Summary

`GET /api/notifications` runs a single unbounded `findMany` — no `take`, no cursor, no expiry filter. Every notification a user has ever received is serialised on every page load of the header bell. There is also no way to delete a notification or to mark them all read, so the row count only ever grows.

### Where

`src/app/api/notifications/route.ts`

```ts
const [notifications, unreadCount] = await Promise.all([
  prisma.notification.findMany({
    where: { userId: session.user.id },   // no take, no cursor, no expiresAt filter
    orderBy: { createdAt: "desc" },
  }),
  prisma.notification.count({ where: { userId: session.user.id, read: false } }),
]);
```

### Why this is a problem

1. **Unbounded payload.** `NotificationBell` fetches this on mount for every authenticated page. An active user who has received a few thousand connection requests / message notifications downloads all of them to render a dropdown that is `max-h-96` and shows maybe six.
2. **Expired notifications are still served.** The schema has `expiresAt` and `src/lib/notifications.ts` has `cleanupExpiredNotifications()`, but that runs on a cron (`/api/internal/notifications/cleanup`) in batches of ≤500. Anything expired but not yet swept is still returned to the client and still rendered. The endpoint should not depend on the sweeper having caught up.
3. **The rest of the API already solved this.** `src/lib/pagination.ts` provides `parsePaginationParams` / `buildTimestampCursorWhere` / `createPaginatedResponse`, and `/api/messages`, `/api/routes`, `/api/tickets` and `/api/users` all use it. Notifications is the last unpaginated list endpoint.
4. **No dismiss, no mark-all-read.** There is `PATCH /api/notifications/[id]/read` and a duplicate `PATCH /api/notifications/[id]` that does the same thing, but no `DELETE`. `markAllNotificationsAsRead()` is exported from `src/lib/notifications.ts` and has **no caller anywhere in the codebase** — the helper was written but never wired to a route. So the only way to clear a 500-item bell is to click every row.

### Steps to reproduce

1. Seed a user with a few hundred `Notification` rows, at least one with `expiresAt` in the past.
2. Sign in and open any page.
3. Network tab: `GET /api/notifications` returns all rows in one response, including the expired one.
4. There is no request the UI can make to dismiss a notification or clear the list.

### Expected

- `GET /api/notifications` is cursor-paginated with the shared helpers, defaulting to 20, capped at 100, returning `pagination.nextCursor` / `pagination.hasMore`.
- Rows whose `expiresAt` has passed are excluded from both the list and `unreadCount`.
- An optional `?unreadOnly=true` filter for the bell's default view.
- `PATCH /api/notifications` marks everything read (wires up the existing `markAllNotificationsAsRead`).
- `DELETE /api/notifications/[id]` dismisses one; `DELETE /api/notifications?read=true` clears the read ones.
- `NotificationBell` gets "Mark all read", per-row dismiss, and a "Load more" control.

MOHITKOURAV01:
Closes #381

## The problem

`GET /api/notifications` was a single unbounded query:

```ts
prisma.notification.findMany({
  where: { userId: session.user.id },   // no take, no cursor, no expiry filter
  orderBy: { createdAt: "desc" },
})
```

`NotificationBell` fetches this on mount for every authenticated page, so an active user downloads every notification they have ever received in order to render a dropdown that is `max-h-96` and shows about six of them.

It also returned rows whose `expiresAt` had already passed. The schema has the column and `cleanupExpiredNotifications()` sweeps it, but on a cron in batches of ≤500 — so between sweeps there are expired rows in the table and the read path was handing them to the client.

And there was no way to remove anything. `PATCH /api/notifications/[id]/read` existed; `DELETE` did not, and `markAllNotificationsAsRead()` was exported from `src/lib/notifications.ts` with **no caller anywhere in the codebase**. The helper was written, the route never was, so clearing a 500-item bell meant clicking 500 rows.

### Pagination

Uses the shared helpers from `src/lib/pagination.ts` — the same ones `/api/messages`, `/api/routes`, `/api/tickets` and `/api/users` already use. Default 20, capped at 100, `pagination.nextCursor` / `hasMore` in the response, `400` on a malformed `limit` or `cursor`. The `notifications` key is kept as an alias for `items` so nothing breaks.

### Expiry

`activeNotificationWhere()` filters both the page and `unreadCount`, off a **single clock reading** so the badge and the list cannot disagree about a notification expiring mid-request. It uses `gt: now` against the sweeper's `lte: now`, so the two rules partition the table: a row is either sweepable or visible, never both and never neither.

### The missing write paths

| Route | Does |
| --- | --- |
| `PATCH /api/notifications` | Mark all read — wires up the orphaned `markAllNotificationsAsRead()` |
| `DELETE /api/notifications/[id]` | Dismiss one |
| `DELETE /api/notifications?all=true` | Clear read ones, or everything |

The single-row delete goes through `deleteMany` with the `userId` in the same `where`, so a guessed id belonging to somebody else deletes nothing and returns 404 rather than throwing `RecordNotFound`.

### UI

`NotificationBell` gets **Mark all read**, a per-row **dismiss**, and **Load more**. Both mutations are optimistic with a rollback on failure. The loading state renders skeletons instead of "No notifications yet" while the first page is still in flight — previously the dropdown claimed the list was empty during load.

## Tests

45 new tests across three files:

- `src/lib/__tests__/notification-list.test.ts` — the new helpers, including that the cursor filter and expiry filter both survive under `AND` rather than one replacing the other, and that `deleteNotification` reports 0 instead of throwing for a foreign id
- `src/app/api/notifications/__tests__/route.test.ts` — GET/PATCH/DELETE, including the single-clock-reading assertion
- `src/app/api/notifications/[id]/__tests__/route.test.ts` — the new DELETE

## Verification

- `npx vitest run src/app/api/notifications src/lib/__tests__/notification-list.test.ts src/lib/__tests__/notifications.test.ts` — 45 passed

## Note on the base branch

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -304,6 +304,78 @@ channel and to each participant's personal channel.
 List the caller's conversations, most recently updated first, each with the
 other participant and the latest message for the sidebar.
 
+## Notification endpoints
+
+### `GET /api/notifications`
+
+Return one page of the caller's notifications, newest first.
+
+Query parameters:
+
+| Name | Default | Notes |
+| --- | --- | --- |
+| `limit` | `20` | Capped at `100`. |
+| `cursor` | — | Opaque cursor from a previous `pagination.nextCursor`. |
+| `unreadOnly` | `false` | `true` returns only unread notifications. |
+
+Response:
+
+```json
+{
+  "items": [
+    {
+      "id": "ntf-3",
+      "title": "New connection request",
+      "content": "Asha wants to connect",
+      "link": "/dashboard/connections",
+      "read": false,
+      "createdAt": "2026-08-10T09:15:00.000Z"
+    }
+  ],
+  "pagination": {
+    "limit": 20,
+    "nextCursor": "eyJ2ZXJzaW9uIjoxLCJ0aW1lc3RhbXAiOiIuLi4iLCJpZCI6Im50Zi0yIn0",
+    "hasMore": true
+  },
+  "notifications": [],
+  "unreadCount": 4
+}
+```
+
+`notifications` is an alias for `items`, kept for existing consumers.
+
+Notifications whose `expiresAt` has passed are excluded from both the page and
+`unreadCount`. The cron sweeper at `/api/internal/notifications/cleanup` deletes
+them in batches, so read paths cannot assume it has caught up.
+
+`400` is returned for a malformed `limit` or `cursor`.
+
+### `PATCH /api/notifications`
+
+Mark every unread notification as read.
+
+```json
+{ "ok": true, "updated": 12 }
+```
+
+### `DELETE /api/notifications`
+
+Clear the caller's read notifications. Pass `?all=true` to clear unread ones too.
+
+```json
+{ "ok": true, "deleted": 12 }
+```
+
+### `PATCH /api/notifications/[id]` and `PATCH /api/notifications/[id]/read`
+
+Mark a single notification as read. Both routes are equivalent; the `/read`
+form is the one the notification bell uses.
+
+### `DELETE /api/notifications/[id]`
+
+Dismiss a single notification. Responds `404` when the id does not exist or
+belongs to another user.
+
 ## AI chat endpoint
 
 ### `POST /api/chat`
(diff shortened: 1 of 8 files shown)


MOHITKOURAV01:

singh-odyssey (approved):

PR #387, opened 2026-08-10
fix(meetup): validate checklist writes and allow items to be deleted

### LINKED ISSUE #382: [Bug] Meetup checklist API writes unvalidated input to the database and has no DELETE — items cannot be removed
### Summary

`/api/meetup-checklist` is the only write endpoint in the app that reads `await req.json()` straight into Prisma with no schema. `text` is never validated — not for type, not for emptiness, not for length. And there is no `DELETE` handler at all, so a checklist item added by mistake is permanent for both travellers.

### Where

`src/app/api/meetup-checklist/route.ts`

```ts
export async function POST(req: NextRequest) {
  ...
  const body = await req.json();          // untyped

  if (!body.meetupPlanId) { ... }         // only meetupPlanId is checked

  const item = await prisma.meetupChecklistItem.create({
    data: {
      meetupPlanId: body.meetupPlanId,
      text: body.text,                    // never validated
    },
  });
```

```ts
export async function PATCH(req: NextRequest) {
  ...
  const existing = await prisma.meetupChecklistItem.findUnique({
    where: { id: body.id },               // body.id may be undefined -> throws
    ...
  });
```

### Why this is a problem

1. **Empty items.** `POST { meetupPlanId, text: "" }` creates a blank row. `TripBoard` guards with `if (!text.trim()) return`, but that is client-side only — curl or a double-submit gets straight through. Same for `PATCH { id, text: "" }`, which blanks an existing item.
2. **No length cap.** `text` is an unbounded `String` in the schema. A 5 MB string is accepted and then shipped to the other traveller on every checklist load.
3. **Wrong types reach Prisma.** `text: 123` or `text: { }` produces an unhandled `PrismaClientValidationError`, which surfaces as a 500 with a stack trace in the logs rather than a 400. `POST` and `PATCH` have no try/catch at all, so a malformed `req.json()` body also 500s.
4. **`PATCH` with no `id`.** `findUnique({ where: { id: undefined } })` throws before the ownership check runs — again a 500 instead of a 400.
5. **No unbounded-growth guard.** Nothing stops a participant adding 100k items to a shared plan; `GET` then returns all of them with no `take`.
6. **No delete.** `TripBoard.tsx` has add, toggle and inline edit, but no remove, because the API has no `DELETE`. The comparable resources (`/api/routes`, `/api/notifications` after #381) all support removal.

Every other route in the repo already uses `withValidation` + a zod schema (`/api/meetup-plans`, `/api/user/report`, `/api/routes`, `/api/tickets`, `/api/chat`). This one was missed.

### Steps to reproduce

```bash
# blank item is accepted
curl -b <cookie> -X POST localhost:3000/api/meetup-checklist \
  -H 'content-type: application/json' \
  -d '{"meetupPlanId":"<id>","text":""}'          # 200, creates an empty row

# non-string text 500s
curl -b <cookie> -X POST localhost:3000/api/meetup-checklist \
  -H 'content-type: application/json' \
  -d '{"meetupPlanId":"<id>","text":{"a":1}}'     # 500 PrismaClientValidationError

# PATCH without an id 500s
curl -b <cookie> -X PATCH localhost:3000/api/meetup-checklist \
  -H 'content-type: application/json' -d '{}'     # 500
```

### Expected

- zod schemas for `POST`, `PATCH` and a new `DELETE`, with `text` trimmed, `min(1)` and `max(200)`.
- Malformed JSON and validation failures return 400, not 500.
- A per-plan item cap (200) returning 409 when exceeded, and a `take` on `GET`.
- `DELETE /api/meetup-checklist?id=<itemId>` gated by the same `canAccessMeetupPlan` check as the other verbs.
- `TripBoard` gets a delete control with optimistic removal and rollback on failure.

MOHITKOURAV01:
Closes #382

## The problem

`/api/meetup-checklist` was the only write endpoint in the app reading `await req.json()` straight into Prisma with no schema:

```ts
const body = await req.json();

if (!body.meetupPlanId) { ... }      // the only thing checked

const item = await prisma.meetupChecklistItem.create({
  data: {
    meetupPlanId: body.meetupPlanId,
    text: body.text,                  // never validated
  },
});
```

What got through:

| Request | Result on `main` |
| --- | --- |
| `POST { meetupPlanId, text: "" }` | 200, blank row created |
| `PATCH { id, text: "   " }` | 200, existing item blanked |
| `POST { meetupPlanId, text: {a:1} }` | 500 `PrismaClientValidationError` |
| `PATCH {}` | 500 — `findUnique({ where: { id: undefined } })` throws *before* the ownership check |
| `POST { meetupPlanId, text: <5MB> }` | 200 — `text` is an unbounded `String`, and both travellers download it on every load |

`TripBoard` guards with `if (!text.trim()) return`, but that is client-side only; curl or a double-submit walks past it. Neither `POST` nor `PATCH` had a `try/catch`, so a malformed body also 500'd.

There was also **no `DELETE` at all**. The list is shared between two travellers and had add, toggle and inline edit but no way to remove anything, because the API had no delete verb.

## What this changes

**Validation.** zod schemas on all three verbs, via `withValidation` — the same pattern `/api/meetup-plans`, `/api/user/report`, `/api/routes`, `/api/tickets` and `/api/chat` already use. `text` is trimmed, required, capped at 200 characters. `completed` must actually be a boolean. A `PATCH` carrying neither field is a 400 rather than a silent no-op update. Malformed JSON is a 400, not a 500.

**Delete.** `DELETE /api/meetup-checklist?id=<itemId>`, gated by the same conversation-membership rule as the other verbs. That check is factored into `loadAccessibleItem()` so every verb that works on an existing item applies it identically, and so `PATCH` resolves the item *after* validating that an id was supplied.

**Growth.** A plan caps at 200 items (409 past that) and `GET` gets a matching `take`, so one participant cannot grow a shared list without bound.

**UI.** `TripBoard` gets a per-row remove button with optimistic removal and rollback, `maxLength` on both inputs, and an error banner surfacing what the server actually said — previously every failure path was a bare `if (!res.ok) return`, so a rejected write looked identical to nothing happening.

## Tests

36 new tests:

- `src/app/api/meetup-checklist/__tests__/route.test.ts` (29) — every rejection above, plus the access-control matrix for all four verbs
- `src/components/__tests__/trip-board.test.tsx` (7) — the delete flow, the rollback, the trimmed payload, the error banner

### One thing worth flagging for reviewers

The component test stubs `lucide-react`. Its ESM build does not resolve under this repo's vitest/jsdom transform — **every** named icon comes back `undefined`, so any component that renders one fails to mount with "Element type is invalid". That is pre-existing and affects any future component test in this repo, not something this PR introduced. Worth fixing properly at the vitest config level; I've kept it to a local stub here rather than widen the scope.

## Verification

- `npx vitest run src/app/api/meetup-checklist` — 29 passed
- `npx vitest run src/components/__tests__/trip-board.test.tsx` — 7 passed

## Note on the base branch

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -241,6 +241,57 @@ Response:
 }
 ```
 
+## Meetup checklist endpoints
+
+Every verb resolves the item's `MeetupPlan`, then its `Conversation`, and
+requires the caller to be a member of that conversation. Either traveller on a
+plan can read and edit the shared list; nobody outside it can.
+
+### `GET /api/meetup-checklist?meetupPlanId=<id>`
+
+Return the plan's checklist, incomplete items first. Bounded at 200 rows.
+Responds `[]` when `meetupPlanId` is omitted, `403` when the caller is not a
+member of the plan's conversation.
+
+### `POST /api/meetup-checklist`
+
+Add an item.
+
+```json
+{
+  "meetupPlanId": "plan-1",
+  "text": "Book the hostel"
+}
+```
+
+- `text` is trimmed, required, and capped at 200 characters.
+- `201 Created` with the new item.
+- `400 Bad Request` for empty, over-long or non-string text.
+- `409 Conflict` once the plan already holds 200 items.
+
+### `PATCH /api/meetup-checklist`
+
+Update an item's text, its completion flag, or both. At least one of the two
+must be present.
+
+```json
+{
+  "id": "item-1",
+  "text": "Book the hostel",
+  "completed": true
+}
+```
+
+- `400 Bad Request` when `id` is missing, when neither field is supplied, when
+  `text` trims to empty, or when `completed` is not a boolean.
+- `404 Not Found` for an unknown id, `403 Forbidden` for another
+  conversation's item.
+
+### `DELETE /api/meetup-checklist?id=<itemId>`
+
+Remove one item. `400` without an `id`, `404` for an unknown id, `403` for
+another conversation's item.
+
 ## Messaging endpoints
 
 ### `GET /api/messages?conversationId=<id>`
--- a/src/app/api/meetup-checklist/__tests__/route.test.ts
+++ b/src/app/api/meetup-checklist/__tests__/route.test.ts
@@ -1,0 +1,473 @@
+import { NextRequest } from "next/server";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import { auth } from "@/lib/auth";
+import prisma from "@/lib/prisma";
+import {
+  DELETE,
+  GET,
+  MAX_CHECKLIST_ITEMS_PER_PLAN,
+  MAX_CHECKLIST_TEXT_LENGTH,
+  PATCH,
+  POST,
+} from "../route";
+
+vi.mock("@/lib/prisma", () => ({
+  default: {
+    meetupPlan: {
+      findFirst: vi.fn(),
+    },
+    meetupChecklistItem: {
+      findMany: vi.fn(),
+      findUnique: vi.fn(),
+      count: vi.fn(),
+      create: vi.fn(),
+      update: vi.fn(),
+      delete: vi.fn(),
+    },
+  },
+}));
+
+vi.mock("@/lib/auth", () => ({
+  auth: vi.fn(),
+}));
+
+function jsonRequest(
+  method: "POST" | "PATCH",
+  body: unknown,
+) {
+  return new NextRequest(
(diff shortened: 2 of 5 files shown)


MOHITKOURAV01:

Test suite on this branch: no regressions against the `main` baseline of 9 failing files / 23 failing tests. The 36 new tests here (29 API + 7 component) all pass.

singh-odyssey (approved):

singh-odyssey (approved):

PR #388, opened 2026-08-10
fix(connections): apply the block filter to GET and paginate the list

### LINKED ISSUE #383: [Bug] GET /api/connections still lists blocked users and runs three unbounded queries
### Summary

`GET /api/connections` never consults `src/lib/blocking.ts`. Blocking someone removes them from your conversation sidebar (`/api/conversations` filters correctly) but leaves them sitting in your connections list, and leaves you in theirs. The same handler also fires three unbounded `findMany` calls with no pagination.

### Where

`src/app/api/connections/route.ts`

```ts
export async function GET(req: NextRequest) {
  ...
  const incoming = await prisma.connectionRequest.findMany({ where: { receiverId: userId, status: "PENDING" }, ... });
  const outgoing = await prisma.connectionRequest.findMany({ where: { senderId: userId, status: "PENDING" }, ... });
  const accepted = await prisma.connectionRequest.findMany({
    where: { status: "ACCEPTED", OR: [{ senderId: userId }, { receiverId: userId }] },
    ...
  });
```

No `getBlockedUserIds()`, no `take`, no cursor.

### Why this is a problem

1. **The block is half-applied.** `POST` does the right thing — `isBlockedBetween()` guards `send` and `accept`. `GET` does not, so after a block:
   - the blocked user still appears under "Connections" with name, photo, bio and location;
   - a pending request from someone you blocked still sits in your incoming list waiting to be accepted (and accepting it correctly 403s, which just looks broken);
   - the person who blocked you still shows in *your* list, since blocking is symmetric here.
   The sidebar hides the thread but the connections page keeps rendering the person, which defeats the point of the feature and is confusing on both sides.
2. **Inconsistent with `/api/conversations`,** which resolves the block set once with `getBlockedUserIds()` and excludes matching rows. The helper was written to be reused for exactly this and this endpoint predates it.
3. **Three unbounded queries, run serially.** Each `await` blocks the next even though none depends on the others. A user with a large network downloads every accepted connection plus every pending request on each dashboard load, with `include`d user records on all three.
4. `orderBy: { updatedAt: "desc" }` on `accepted` with no tiebreaker is not a stable sort — rows updated in the same transaction can reorder between requests, which will matter as soon as pagination exists.

### Steps to reproduce

1. Users A and B are connected (`ACCEPTED`).
2. A blocks B via `POST /api/user/block`.
3. `GET /api/conversations` as A — the thread with B is correctly gone.
4. `GET /api/connections` as A — B is still in `connections`, with profile fields.
5. `GET /api/connections` as B — A is still in `connections` too.

### Expected

- Resolve `getBlockedUserIds(userId)` once and exclude those ids from `incoming`, `outgoing` and `connections`.
- Run the three queries concurrently instead of serially.
- Cursor-paginate the accepted-connections list with the shared `src/lib/pagination.ts` helpers, and bound the two pending lists.
- Stable ordering: `[{ updatedAt: "desc" }, { id: "desc" }]`.

MOHITKOURAV01:
Closes #383

## The problem

Blocking is only half-applied on this endpoint. `POST` does the right thing — `isBlockedBetween()` guards both `send` and `accept`. `GET` never consulted `src/lib/blocking.ts` at all.

So after A blocks B:

- `GET /api/conversations` correctly drops the thread
- `GET /api/connections` still returns B under `connections`, with name, photo, bio and location
- a pending request from B still sits in A's `incoming` list, and accepting it 403s — which just looks broken
- B still sees A in *their* list too, because blocking is symmetric here

The sidebar hides the person while the connections page keeps rendering them, which defeats the point of the feature and is confusing on both sides.

The same handler also ran three unbounded `findMany` calls **serially**, each with `include`d user records, on every dashboard load.

## What this changes

**Block filter.** `getBlockedUserIds()` resolved once, applied to all three lists. For the accepted list the filter has to sit inside each `OR` branch, because the counterpart is the *receiver* when the caller sent the request and the *sender* when they received it:

```ts
OR: [
  { senderId: userId,   receiverId: { notIn: blockedUserIds } },
  { receiverId: userId, senderId:   { notIn: blockedUserIds } },
]
```

**Concurrency.** The three queries don't depend on each other, so they run in one `Promise.all` instead of each `await` blocking the next.

**Bounds.** The two pending lists cap at 100. The accepted list is cursor-paginated with the shared `src/lib/pagination.ts` helpers.

**Stable ordering.** `orderBy: { updatedAt: "desc" }` alone is not a stable sort — two rows accepted in the same transaction can swap places between requests, which would make the new cursor skip or repeat a row. Adds an `id` tiebreaker, matching what `/api/messages` and `/api/routes` do.

## Backwards compatibility

`incoming`, `outgoing` and `connections` keep their shapes and their keys, so `ChatInterface` and `DashboardMatches` need no changes. `pagination` is additive — a follow-up can wire a "load more" onto it.

## Tests

The existing file was for `GET` (2 cases) and `POST` (4 cases). The `POST` cases are unchanged; the two `GET` cases now pass a real `NextRequest`, since the handler reads `searchParams`. 14 new `GET` cases on top.

Worth noting: the block tests drive `prisma.block.findMany` rather than stubbing `getBlockedUserIds`, so the real symmetry logic stays under test — `blockedWith()` alternates the direction of each `Block` row on purpose.

The concurrency test holds the first query's promise open and asserts all three have started, which fails against the serial version.

## Verification

- `npx vitest run src/app/api/connections` — 20 passed (4 pre-existing POST cases + 16)
- `npx tsc --noEmit` reports nothing new for this path

## Note on the base branch

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -220,6 +220,65 @@ Fetch a single route belonging to the authenticated user.
 
 Delete a saved route owned by the authenticated user.
 
+## Connection endpoints
+
+### `GET /api/connections`
+
+Return the caller's pending requests in both directions plus a page of their
+accepted connections.
+
+Query parameters:
+
+| Name | Default | Notes |
+| --- | --- | --- |
+| `limit` | `20` | Page size for `connections`. Capped at `100`. |
+| `cursor` | — | Opaque cursor from a previous `pagination.nextCursor`. |
+
+Response:
+
+```json
+{
+  "incoming": [],
+  "outgoing": [],
+  "connections": [
+    {
+      "requestId": "req-1",
+      "user": { "id": "user-2", "name": "Asha", "image": null, "bio": null, "location": "Delhi" },
+      "connectedAt": "2026-08-01T12:00:00.000Z"
+    }
+  ],
+  "pagination": {
+    "limit": 20,
+    "nextCursor": "eyJ2ZXJzaW9uIjox...",
+    "hasMore": false
+  }
+}
+```
+
+`incoming` and `outgoing` are the pending requests, newest first, bounded at
+100 each. `connections` is the cursor-paginated accepted list.
+
+Users on either side of a `Block` are excluded from all three lists, matching
+the behaviour of `GET /api/conversations` and the guards `POST
+/api/connections` already applies to `send` and `accept`.
+
+`400` is returned for a malformed `limit` or `cursor`.
+
+### `POST /api/connections`
+
+Send, accept or decline a connection request.
+
+```json
+{
+  "action": "send",
+  "userId": "user-2"
+}
+```
+
+`action` is one of `send`, `accept`, `decline`. Accepting creates the
+conversation between the two users. All three are rate limited and refuse to
+act across a block.
+
 ## Match discovery endpoint
 
 ### `GET /api/matches?destination=<name>&date=<YYYY-MM-DD>`
--- a/src/app/api/connections/__tests__/route.test.ts
+++ b/src/app/api/connections/__tests__/route.test.ts
@@ -1,3 +1,4 @@
+import { NextRequest } from "next/server";
 import { describe, it, expect, beforeEach, vi } from "vitest";
 import prisma from "@/lib/prisma";
 import { auth } from "@/lib/auth";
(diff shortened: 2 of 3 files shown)


MOHITKOURAV01:

singh-odyssey (approved):

Issue #372, opened 2026-08-09
[Security] Missing rate limiting on password reset confirmation endpoint (/api/auth/reset-password)


Pratyush-Panda-2006:
### Summary
The password reset endpoint ([`src/app/api/auth/reset-password/route.ts`](file:///d:/ECSOC/Travellers/travellers/src/app/api/auth/reset-password/route.ts)) accepts a `token` and new `password` to update user credentials. Unlike other sensitive authentication routes (`signin`, `signup`, `verify-otp`, `resend-otp`, `forgot-password`), `/api/auth/reset-password` does not enforce rate limits using `@/lib/rate-limit-rules`.

Unauthenticated attackers can make unlimited requests to brute-force reset tokens or overload the database with password reset attempts.

### Steps to Reproduce
1. Send high-frequency POST requests to `/api/auth/reset-password` with guessed tokens.
2. Observe that none of the requests are throttled with HTTP 429 Too Many Requests.

### Proposed Fix
Define an `authResetPassword` rate limiting rule in [`src/lib/rate-limit-rules.ts`](file:///d:/ECSOC/Travellers/travellers/src/lib/rate-limit-rules.ts) and call `enforceRateLimit(request, "authResetPassword")` inside `/api/auth/reset-password/route.ts`.

Pratyush-Panda-2006:
@singh-odyssey May you please assign me this issue

PR #389, opened 2026-08-10
fix(auth): stop change-password weakening the hash, and rate limit it

### LINKED ISSUE #384: [Security] Change-password endpoint downgrades bcrypt cost to 10 and has no rate limiting
### Summary

`PUT /api/auth/change-password` hashes the new password with a hardcoded `10` bcrypt rounds while signup and reset both use `12` in production, and it is the only credential endpoint in the app with no rate limit. It also accepts a soft-deleted account and happily "changes" a password to the one already in use.

### Where

`src/app/api/auth/change-password/route.ts`

```ts
// Hash the new password (10 rounds)
const newPasswordHash = await hash(newPassword, 10);
```

Compare with `src/app/api/auth/signup/route.ts` and `src/app/api/auth/reset-password/route.ts`:

```ts
const saltRounds = process.env.NODE_ENV === "production" ? 12 : 10;
```

### Why this is a problem

1. **Silent security downgrade.** The cost factor was deliberately chosen — `package.json` still carries a `benchmark:bcrypt` script from when it was measured (the script file itself is no longer in the repo), and both signup and reset-password run 12 in production. Any user who changes their password in prod is silently moved from cost 12 down to cost 10, which is ~4x cheaper to attack offline. Nothing surfaces this; the value is duplicated as a literal in three files and this one drifted.
2. **No rate limiting.** `src/lib/rate-limit-rules.ts` covers signup, signin, verify-otp, resend-otp, forgot-password, chat, messages, connections and reports. `change-password` is absent, so `currentPassword` can be brute-forced from an authenticated session at whatever rate the server will take. It is a password oracle sitting behind a stolen or borrowed session cookie, and it is cheaper to attack than `/api/auth/signin` because that one *is* throttled.
3. **Soft-deleted accounts can still rotate credentials.** The lookup is `findUnique({ where: { email } })` with no `isDeleted: false`, so an account that went through `deleteUserAccount()` can still have its password changed.
4. **No-op changes are accepted.** Setting `newPassword` equal to `currentPassword` returns `{ ok: true, message: "Password updated successfully" }` and rewrites the same secret. If a user is rotating a password because it leaked, the endpoint should say so rather than report success.
5. **Timing/enumeration nit.** When `passwordHash` is null (OAuth account) the handler returns immediately without ever calling `compare`, so the response is measurably faster than the wrong-password path.

### Steps to reproduce

1. In production mode, sign up — the stored hash starts `$2a$12$`.
2. `PUT /api/auth/change-password` with a valid current password.
3. The stored hash now starts `$2a$10$`.
4. Send the same request 500 times with different `currentPassword` values — every one is processed; no 429, no `Retry-After`, no backoff.

### Expected

- One shared helper (e.g. `src/lib/password.ts`) exporting the cost factor and `hashPassword()` / `verifyPassword()`, used by change-password and reset-password so the value cannot drift again.
- A `authChangePassword` rule in `RATE_LIMIT_RULES`, applied with `enforceRateLimit` and returning 429 + `Retry-After` like the other auth routes.
- `isDeleted: false` in the user lookup.
- Reject `newPassword === currentPassword` with a 400.
- Always run a bcrypt comparison so the OAuth-account path costs the same as a wrong password.

### Note

This is a different endpoint from #372, which covers `POST /api/auth/reset-password`. This one is the authenticated change-password route.

MOHITKOURAV01:
Closes #384

> Different endpoint from #372, which is about `POST /api/auth/reset-password`. This one is the authenticated change-password route.

### 1. The cost factor drifted

`src/app/api/auth/signup/route.ts` and `src/app/api/auth/reset-password/route.ts`:

```ts
const saltRounds = process.env.NODE_ENV === "production" ? 12 : 10;
```

`src/app/api/auth/change-password/route.ts`:

```ts
// Hash the new password (10 rounds)
const newPasswordHash = await hash(newPassword, 10);
```

The same value written as a literal in three places, and one of them drifted. **Every password change in production silently moved that user's hash from cost 12 down to cost 10** — roughly 4x cheaper to attack offline — and nothing surfaced it.

### 2. No rate limit

`RATE_LIMIT_RULES` covers signup, signin, verify-otp, resend-otp, forgot-password, chat, messages, connections and reports. `change-password` was absent.

The endpoint takes `currentPassword`, so an unthrottled version is a password oracle for anyone holding a session cookie — and a **cheaper** one than `/api/auth/signin` next door, because that one is throttled. Send it 500 times with different guesses and all 500 are processed.

### 3. Three smaller things

- The lookup had no `isDeleted` filter, so an account that had been through `deleteUserAccount()` could still rotate its credentials.
- `newPassword === currentPassword` returned `{ ok: true, message: "Password updated successfully" }` and rewrote the same secret. For somebody rotating a password *because it leaked*, that is an actively misleading success.
- The OAuth-account branch returned before any bcrypt call, so it answered measurably faster than a wrong password.

## What this changes

**`src/lib/password.ts`** — the cost factor and `hashPassword` / `verifyPassword` in one place. `reset-password` uses it too, so the value cannot drift again.

`signup` is deliberately **not** touched: it does not currently compile on `main` and is being edited by #379. It should adopt the helper once that lands — I've left it alone rather than create a conflict.

**Rate limit** — `authChangePassword`, 5 per 15 minutes, keyed on the account rather than the IP alone so one user's mistyping cannot lock another out from a shared connection. `X-RateLimit-*` headers on the success and failure paths alike, `Retry-After` on the 429. The check runs before the DB read, so a throttled request costs no query.

**`verifyOptionalPassword()`** spends a comparison against a cached dummy hash when `passwordHash` is null, so the OAuth branch costs the same as a wrong password. The dummy is computed once per process, not per request.

Plus the `isDeleted` filter and the same-password rejection.

## Tests

26 new tests:

- `src/lib/__tests__/password.test.ts` (11) — cost factor per environment, that the cost is encoded in the emitted hash, that hashes are salted, that a hash written at cost 12 still verifies after the factor changes, and that the no-hash path actually spends a comparison
- `src/app/api/auth/change-password/__tests__/route.test.ts` (15) — the 429 path costing no query, headers on failure paths, the `isDeleted` filter, same-password rejection, and that the stored value is a real bcrypt hash

### Test consolidation — please look at this bit

`src/app/api/auth/__tests__/change-password.test.ts` is **removed**, and its five cases are folded into `src/app/api/auth/change-password/__tests__/route.test.ts` — which is where every other route in this repo keeps its tests.

Nothing it covered is lost. 401, empty input, short password, OAuth account, wrong current password and the successful update are all asserted in the new file, and against a real bcrypt hash rather than a stubbed `bcryptjs` module (the old file mocked `hash` to return the literal string `"new_hashed_password"`, which could not have caught the cost-factor bug this PR fixes). I went this way rather than leaving two files testing the same route.

## Verification

- `npx vitest run src/lib/__tests__/password.test.ts src/app/api/auth/change-password` — 26 passed

## Note on the base branch

`main` does not currently typecheck (#366, fixed by #379). Nothing here touches those five files.


MOHITKOURAV01:

singh-odyssey (approved):

PR #397, opened 2026-08-12
fix: green the test suite and run vitest in CI (#318)

### LINKED ISSUE #318: Test suite is red (3 failing) and CI never runs vitest

vedant7007:
Closes #318

### Problem
The test suite was red and CI never ran vitest, so regressions could land unnoticed.

### Root causes & fixes
**Auth suites (signup / signin / forgot-password)** — the real cause of the red tests:
- They did `require("@/lib/rate-limit")` at runtime, but the `@` path alias is only resolved at transform time (via `vite-tsconfig-paths`), so it threw `Cannot find module`. Replaced with a namespace `import * as rateLimit from "@/lib/rate-limit"` of the (mocked) module.
- Added the missing `hashOTP` mock (the signup route imports it).
- The `rateLimitExceededResponse` mocks returned a plain `{status:429}` object; `withValidation` then calls `withRequestId(response)` which sets a header → threw (500). Return real `NextResponse` objects.

**tickets test** — the route now checks for a duplicate before creating, so it calls `prisma.ticket.findMany` (missing from the mock → 500). Added it, and updated the assertion to `201 Created` (the route’s actual, correct status).

**ticket-upload-form** — the icons from `lucide-react` resolve to `undefined` under vitest’s ESM resolution, so rendering threw "Element type is invalid". Stubbed them in the test. Also fixed two real gaps in the component that the tests (correctly) rely on:
- `aria-label` on the `<form>` and `htmlFor` on the file `<label>` (a11y — and makes `getByRole("form")` / `getByLabelText` work).
- A **synchronous `submittingRef` guard** so rapid repeated submits fire the request only once — the `loading` state is async and can’t prevent synchronous re-entry (the test fired 3 submits → 3 requests).

**CI** — added a `Test` step (`npm test`) so vitest runs on every push/PR.

### Result
**All 329 tests pass across 52 files**, and CI now runs the suite.

---

**Follow-up (CI):** the new Test step first failed on CI with `ERR_REQUIRE_ESM` — the workflow pinned Node **18.18.0**, which is EOL and can't `require()` the now ESM-only `vite` that vitest 3 loads. Bumped the workflow to Node **22** (current LTS, matches where the suite is validated). Lint and Type check were already green on this branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,8 @@ jobs:
         run: npm run lint -- --no-cache
       - name: Type check
         run: npx tsc --noEmit
+      - name: Test
+        run: npm test
       - name: Build
-        
+
         run: npx next build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '18.18.0'
+          node-version: '22'
           cache: 'npm'
       - name: Show versions
         run: |
--- a/src/app/api/auth/__tests__/forgot-password.test.ts
+++ b/src/app/api/auth/__tests__/forgot-password.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach, vi } from "vitest";
+import { NextResponse } from "next/server";
 
 vi.mock("@/lib/prisma", () => ({
   default: {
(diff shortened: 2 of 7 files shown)


vedant7007:
Hi maintainers 👋 — #397 (closes #318) is green: it fixes the red test suite (all 329 tests pass across 52 files), wires vitest into CI, and bumps the workflow off EOL Node 18 → 22 (which is what let vitest 3 run at all). Ready for review whenever you have a chance — happy to address any feedback. Thanks! 🙏

singh-odyssey (approved):

PR #403, opened 2026-08-16
fix(routes): use findFirst instead of findUnique for compound where clause

### LINKED ISSUE #398: [Bug] GET /api/routes/[id] throws PrismaClientValidationError due to non-unique where constraint
### Summary
In [`src/app/api/routes/[id]/route.ts`](file:///d:/ECSOC/Travellers/travellers/src/app/api/routes/[id]/route.ts#L30-L35), fetching a single route by ID executes:
```ts
const route = await prisma.route.findUnique({
  where: {
    id: params.id,
    userId: session.user.id,
  },
});
```

However, `Route` in [`prisma/schema.prisma`](file:///d:/ECSOC/Travellers/travellers/prisma/schema.prisma#L222) does not have a composite `@@unique([id, userId])` constraint. Prisma Client's `findUnique` strictly requires fields that form a unique identifier.

When this query runs, Prisma throws a `PrismaClientValidationError`, causing the endpoint to fail with an unhandled 500 Internal Server Error (`"Unable to fetch the route"`) instead of returning the route or a 404 response.

### Steps to Reproduce
1. Create and save a route for an authenticated user.
2. Send a `GET /api/routes/<routeId>` request with a valid session.
3. Observe the 500 Internal Server Error response caused by Prisma validation failure.

### Proposed Fix
Change `prisma.route.findUnique` to `prisma.route.findFirst({ where: { id: params.id, userId: session.user.id } })` in `src/app/api/routes/[id]/route.ts`.

Pratyush-Panda-2006:
Fixes #398

### Summary
In \src/app/api/routes/[id]/route.ts\, replaced \prisma.route.findUnique\ with \prisma.route.findFirst\ because \Route\ does not have a composite \@@unique([id, userId])\ constraint in schema. This fixes the \PrismaClientValidationError\ and 500 error when retrieving a route by ID.

### Verification
Added unit tests in \src/app/api/routes/[id]/__tests__/route.test.ts\ verifying 200 formatted response, 401 unauthorized, and 404 not found cases.

### DIFF
--- a/src/app/api/routes/[id]/__tests__/route.test.ts
+++ b/src/app/api/routes/[id]/__tests__/route.test.ts
@@ -1,0 +1,97 @@
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+vi.mock("@/lib/auth", () => ({
+  auth: vi.fn(),
+}));
+
+vi.mock("@/lib/prisma", () => ({
+  default: {
+    route: {
+      findFirst: vi.fn(),
+    },
+  },
+}));
+
+import { GET } from "../route";
+import { auth } from "@/lib/auth";
+import prisma from "@/lib/prisma";
+
+describe("GET /api/routes/[id]", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  const createNextRequest = () => {
+    return {
+      headers: new Headers(),
+    } as any;
+  };
+
+  it("returns 401 if unauthenticated", async () => {
+    (auth as any).mockResolvedValue(null);
+
+    const req = createNextRequest();
+    const res = await GET(req, { params: { id: "route-123" } });
+    const body = await res.json();
+
+    expect(res.status).toBe(401);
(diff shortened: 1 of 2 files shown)


santichausis (approved):
Read the route and the test in full. Correct fix — \`findUnique\` requires the where clause to hit a unique constraint, and \`Route\` has no \`@@unique([id, userId])\`, so \`{id, userId}\` was throwing \`PrismaClientValidationError\`. \`findFirst\` is the right swap. Small, well-tested (401/404/200 all covered).

Note: automated CI (\`build-and-lint\`) shows \`action_required\` on this branch — it never ran, and I only have \`triage\` access so I can't approve the workflow run myself. That's a repo-permission gap, not a problem with this PR; @singh-odyssey would need to approve it. Verified correctness by reading the diff directly instead.

singh-odyssey (approved):

singh-odyssey:
@santichausis 
you now have write access .

PR #406, opened 2026-08-16
fix(user/block): remove all connection requests when blocking a user

### LINKED ISSUE #401: [Bug] Blocking a user leaves existing ACCEPTED connection requests intact
### Summary
In [`src/app/api/user/block/route.ts`](file:///d:/ECSOC/Travellers/travellers/src/app/api/user/block/route.ts#L164-L180), when a user blocks another user, the transaction deletes only connection requests with `status: "PENDING"`:
```ts
prisma.connectionRequest.deleteMany({
  where: {
    status: "PENDING",
    OR: [
      { senderId: session.user.id, receiverId: blockedId },
      { senderId: blockedId, receiverId: session.user.id },
    ],
  },
});
```

However, existing `ACCEPTED` connection requests between the two users are not deleted. As a result, the blocked relationship remains stored as an `ACCEPTED` connection in the `ConnectionRequest` table, creating an inconsistent connection state across user relationship queries.

### Steps to Reproduce
1. Establish an accepted connection between User A and User B.
2. User A blocks User B via `POST /api/user/block`.
3. Inspect the `ConnectionRequest` database table and observe that the `ACCEPTED` connection request record is still present.

### Proposed Fix
Update `prisma.connectionRequest.deleteMany` in `src/app/api/user/block/route.ts` to remove all connection requests between the two users regardless of status (`PENDING` or `ACCEPTED`).

Pratyush-Panda-2006:
Fixes #401

### Summary
In \src/app/api/user/block/route.ts\, updated \prisma.connectionRequest.deleteMany\ within the block transaction to remove all connection requests (both \PENDING\ and \ACCEPTED\) between the two users. This avoids leaving active connection records after a user is blocked.

### Verification
Added unit test in \src/app/api/user/block/__tests__/block-accepted-connection.test.ts\ verifying that all connection records between the users are removed upon blocking.

### DIFF
--- a/src/app/api/user/block/__tests__/block-accepted-connection.test.ts
+++ b/src/app/api/user/block/__tests__/block-accepted-connection.test.ts
@@ -1,0 +1,72 @@
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+vi.mock("@/lib/auth", () => ({
+  auth: vi.fn(),
+}));
+
+vi.mock("@/lib/prisma", () => ({
+  default: {
+    user: {
+      findUnique: vi.fn(),
+    },
+    block: {
+      findUnique: vi.fn(),
+      create: vi.fn(),
+    },
+    connectionRequest: {
+      deleteMany: vi.fn(),
+    },
+    $transaction: vi.fn(async (callback: any) => {
+      if (typeof callback === "function") {
+        return callback(prisma);
+      }
+      return Promise.all(callback);
+    }),
+  },
+}));
+
+import { POST } from "../route";
+import { auth } from "@/lib/auth";
+import prisma from "@/lib/prisma";
+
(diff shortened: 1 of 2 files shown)


santichausis (approved):
Small and correct — read the full diff. Dropping the \`status: "PENDING"\` filter from the \`deleteMany\` where-clause means blocking now clears an ACCEPTED connection too, matching #401. Checked that this doesn't need to also touch \`Conversation\`: blocked counterparts are already filtered out of \`/api/conversations\` at query time (from #432), so a stale conversation row isn't visible to either side anyway.

Same CI caveat as the others in this batch: \`build-and-lint\` is \`action_required\` and I can't approve that run (triage-only access). Verified by reading the code instead.

singh-odyssey (approved):

PR #413, opened 2026-08-19
fix: PATCH /api/user/profile leaks passwordHash + live reset token (#408)

### LINKED ISSUE #408: [Security] PATCH /api/user/profile leaks passwordHash, OTP and live reset token in the response
The profile-update handler returns the **entire** `User` row, including authentication secrets.

### Root cause
`src/app/api/user/profile/route.ts:106-161` — `PATCH` calls:
```ts
const updatedUser = await prisma.user.update({ where: { email }, data: {...} });   // no select / omit
return apiJson(updatedUser, requestId);
```
Prismas `update` returns the full row by default, so the JSON includes `passwordHash`, `otp`, `otpExpires`, `resetToken`, `resetTokenExpires`, `emailVerified`, `role`, `isDeleted`. There is no global mitigation (`src/lib/prisma.ts` sets no `omit`), and the sibling `GET` handler in the same file (lines 26-47) deliberately uses an explicit `select` that excludes these — the `PATCH` path just doesnt.

### Failure scenario
A logged-in user with an in-flight password reset (`resetToken` set by `/api/auth/forgot-password`) edits any profile field (e.g. bio). The 200 response body then contains their bcrypt `passwordHash` **and the still-valid `resetToken`** — a bearer credential that grants account takeover via `/api/auth/reset-password`. That token is now exposed to browser history, client telemetry/logging, any XSS, or an intermediary proxy; the hash additionally enables offline cracking.

### Fix
Add an explicit `select` to the `update` (mirroring the `GET` projection), or configure client-level `omit` for the sensitive columns.

### Dedup
Unreported. Open PRs #390 (socialLinks validation) and #393 (soft-deleted access) edit this file but neither adds a `select` (confirmed by reading #390s diff). No issue mentions this leak.

vedant7007:
Fixes #408

### Problem
`PATCH /api/user/profile` did `prisma.user.update({ where, data })` with **no `select`/`omit`** and returned the result via `apiJson(updatedUser, requestId)`. Prisma’s `update` returns the full row, so the JSON response included `passwordHash`, `otp`, `otpExpires`, `resetToken`, `resetTokenExpires`, `emailVerified`, `role`, `isDeleted`.

A logged-in user with an in-flight password reset who edits any field (e.g. bio) got their bcrypt hash **and still-valid `resetToken`** in the 200 body — an account-takeover credential exposed to browser history, client telemetry, XSS, or proxies. (The sibling `GET` already uses an explicit `select`; `PATCH` didn’t.)

### Fix
Added an explicit `select` to the `update`, mirroring the `GET` projection, so only non-sensitive fields are returned. Updated the one `profile-expansion` assertion that pinned the exact `update()` args.

### Note on CI
Three profile tests fail on **current `main` independently of this change** — `profile-social-links` (×2) and `profile-expansion` "returns profile with rich fields" — because their `prisma.user.findUnique` mock predates the `isDeleted` guard the PATCH/GET handlers now run (they get `404` instead of `200`). I confirmed they fail on a clean checkout without this PR. Happy to fix the suite in a separate PR if you’d like.

### DIFF
--- a/src/app/api/user/__tests__/profile-expansion.test.ts
+++ b/src/app/api/user/__tests__/profile-expansion.test.ts
@@ -172,6 +172,9 @@ const res = await profileGET(
           gender: undefined,
           travelStyle: undefined,
         },
+        // PATCH now returns a non-sensitive projection (excludes passwordHash,
+        // otp, resetToken, ...) — the exact fields aren't asserted here.
+        select: expect.any(Object),
       });
     });
   });
--- a/src/app/api/user/profile/route.ts
+++ b/src/app/api/user/profile/route.ts
@@ -173,6 +173,28 @@ export async function PATCH(request: NextRequest) {
             ? body.travelStyle
             : undefined,
       },
+      // Return only non-sensitive columns — Prisma's default update returns the
+      // whole row (passwordHash, otp, resetToken, ...). Mirror the GET projection.
+      select: {
+        name: true,
+        email: true,
+        bio: true,
+        location: true,
+        image: true,
+        homeLocation: true,
+        phone: true,
+        emailVerified: true,
+        createdAt: true,
+        languages: true,
+        travelInterests: true,
+        accommodationPrefs: true,
+        budgetRange: true,
+        socialLinks: true,
+        age: true,
+        gender: true,
+        travelStyle: true,
+        isDeleted: true,
+      },
     });
 
     return apiJson(updatedUser, requestId);


singh-odyssey (approved):

singh-odyssey:
@vedant7007 
kindly fix build and  lint test

vedant7007:
Gentle nudge on this one 🙏 — heads up that the red `build-and-lint` is **pre-existing** and unrelated to this PR: it fails on `src/lib/rate-limit-rules.ts` and `src/app/api/admin/tickets/[id]/__tests__/cache-invalidation.test.ts`, neither of which this PR touches (it only changes `src/app/api/user/profile/route.ts` + its test). So it should be safe to review/merge. Happy to help fix that repo-wide lint separately if useful.

santichausis:

PR #414, opened 2026-08-20
fix: rate limiting bypassable via spoofed leftmost X-Forwarded-For (#410)

### LINKED ISSUE #410: [Security] Rate limiting bypassable via spoofed leftmost X-Forwarded-For
The client IP is taken from the attacker-controllable end of `X-Forwarded-For`, so per-IP limits can be trivially evaded.

### Root cause
`src/lib/client-ip.ts:38-48` — `getClientIp` returns the **first (leftmost) valid IP** in `X-Forwarded-For` *before* falling back to platform headers (`x-real-ip`, `cf-connecting-ip`, `fly-client-ip`). The leftmost XFF entry is the client-supplied, spoofable end of the chain. `getRateLimitIdentifier` (`src/lib/rate-limit.ts:134-144`) keys every throttle counter on this value.

### Failure scenario
For any per-IP limit — most importantly the unauthenticated `chat` (Gemini cost-abuse) limit, and the IP component of `authSignin` / `authForgotPassword` / `authVerifyOtp` — an attacker rotates `X-Forwarded-For: <random-ip>` per request. Each spoofed value hashes to a fresh bucket, so the counter never accumulates and the limit never trips, reopening the OTP/credential brute-force and Gemini cost-abuse vectors those limits (#360, #319) were added to close. Endpoints that also scope by email/user-id are partially protected; the IP-only `chat` limit is fully bypassable.

_(Severity is deployment-dependent — it matters when the platform preserves rather than overwrites client-supplied XFF.)_

### Fix
Prefer a trusted platform header (`x-real-ip` / `cf-connecting-ip`) over raw `X-Forwarded-For`, or take the rightmost untrusted hop.

### Dedup
Rate-limit issues #285/#352/#360 added the limiter but none addresses IP-source trust. Unreported.

vedant7007:
Fixes #410

### Problem
`getClientIp` (`src/lib/client-ip.ts`) returned the **first (leftmost)** valid `X-Forwarded-For` entry *before* falling back to platform headers. The leftmost entry is the client-controllable end of the chain, so an attacker could rotate `X-Forwarded-For: <random>` per request — each value hashing to a fresh Redis bucket — and never trip the per-IP limits (auth brute-force via `authSignin`/`authForgotPassword`/`authVerifyOtp`, and the unauthenticated `chat` Gemini cost-abuse limit).

### Fix
- **Prefer the platform-set headers first** (`x-real-ip` / `cf-connecting-ip` / `fly-client-ip`) — a CDN/proxy overwrites these, so the caller can’t spoof them.
- Only then fall back to `X-Forwarded-For`, taking the **rightmost** (nearest-hop) valid entry instead of the leftmost.

### Tests
Updated `client-ip.test.ts`: the forwarded case now expects the rightmost address, and a new test asserts platform headers win over a spoofed `X-Forwarded-For`. (4 pass.)

### DIFF
--- a/src/lib/__tests__/client-ip.test.ts
+++ b/src/lib/__tests__/client-ip.test.ts
@@ -10,12 +10,21 @@ function requestWithHeaders(
 }
 
 describe("getClientIp", () => {
-  it("uses the first valid forwarded address", () => {
+  it("uses the rightmost (nearest-hop) forwarded address, not the spoofable leftmost", () => {
     const request = requestWithHeaders({
       "x-forwarded-for": "203.0.113.5, 10.0.0.2",
     });
 
-    expect(getClientIp(request)).toBe("203.0.113.5");
+    expect(getClientIp(request)).toBe("10.0.0.2");
+  });
+
+  it("prefers trusted platform headers over a spoofable X-Forwarded-For", () => {
+    const request = requestWithHeaders({
+      "x-forwarded-for": "1.2.3.4",
+      "x-real-ip": "198.51.100.7",
+    });
+
+    expect(getClientIp(request)).toBe("198.51.100.7");
   });
 
   it("falls back to platform IP headers", () => {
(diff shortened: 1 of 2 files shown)


singh-odyssey (approved):

singh-odyssey:
@vedant7007 
kindly fix lint test

santichausis:

PR #415, opened 2026-08-21
fix: match-cache invalidation misses substring searches (#409)

### LINKED ISSUE #409: [Bug] Match-cache invalidation misses substring searches — stale match results after a ticket is verified
Match search filters by substring, but the cache is invalidated by the tickets full destination string, so partial-search caches are never busted.

### Root cause
- Search filters tickets with `destination: { contains: destination, mode: "insensitive" }` (`src/app/api/matches/route.ts:276-279`), and the per-user cache key is `normalizeMatchDestination(searchQuery)` (`src/lib/match-cache.ts`, `buildMatchCacheKey` → `versionKey`).
- But `invalidateMatchCachesForTicket` bumps the version keyed on the tickets **own full destination** (`normalizeMatchDestination(identity.destination)`, `match-cache.ts:255-263`). The version key for a search query only matches when the query equals the tickets full destination.

### Failure scenario
A ticket to `"Paris, France"` is verified by an admin → increments the version for `"paris, france"`. A user searching `"Paris"` reads/writes cache under version key `"paris"`, which is never incremented → they keep getting stale results that **omit the newly-verified traveller** until the 5-minute TTL expires. Any partial/substring search (the common case) is affected; only exact-string searches invalidate correctly.

### Fix
Invalidate at the substring/normalized-prefix granularity the query uses (or drop the version-key destination scoping and invalidate the users match caches broadly on ticket verification).

### Dedup
Distinct from #320 (100-cap). Prior cache work (#289/#301, #361) added `invalidateMatchCachesForTicket` generally; this substring/exact mismatch is an unaddressed residual gap. Unreported.

vedant7007:
Fixes #409

### Problem
Match search filters tickets by **substring** (`destination: { contains: query }`), but the cache **version** was keyed by the ticket’s **full normalized destination**. A search for `"Paris"` read `matches-version:paris:<date>`, while verifying a `"Paris, France"` ticket only bumped `matches-version:paris, france:<date>` — so the `"Paris"` search kept serving **stale** results (omitting the newly-verified traveller) until the 5-min TTL. Every partial/substring search (the common case) was affected; only exact-string searches invalidated correctly.

### Fix
Key the cache version by **date only**, so verifying a ticket for a date invalidates *every* destination search for that date (substring or not). The cache key itself still includes the destination — only the version dimension drops it. `getMatchCacheVersion` / `incrementVersion` / `invalidateMatchCachesForTicket` are updated accordingly (scoped to the same ±window of affected dates as before).

### Tests
Updated the existing version tests to the date-only signature and added a regression test: a `"Paris, France"` ticket now invalidates a `"Paris"` substring search (`v0 → v1`). It fails on the old lib and passes now (9/9).

### DIFF
--- a/src/lib/__tests__/match-cache.test.ts
+++ b/src/lib/__tests__/match-cache.test.ts
@@ -159,9 +159,7 @@ describe("cache invalidation", () => {
       "2026-08-15",
     )) {
       await expect(
-        getMatchCacheVersion(
-          "goa",
-          date,
+        getMatchCacheVersion(date,
           store,
         ),
       ).resolves.toBe(1);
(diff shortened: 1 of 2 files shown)


singh-odyssey:
@vedant7007 
lint test is failing , possible problems
-  Duplicate function calls and variable declarations creating syntax errors.
- Lines 102 and 116 both declare const tx = transactionMock(...), and lines 129-130 have duplicate/malformed function call syntax.
-  rate-limit-rules.ts (Line 102)

kindly fix that

santichausis:

PR #428, opened 2026-08-26
security(onboard): validate the onboarding payload and guard soft-deleted accounts

### LINKED ISSUE #423: [Security] POST /api/user/onboard accepts a completely unvalidated body and skips the soft-delete check
## Summary

`POST /api/user/onboard` is the only write endpoint on the user object that has **no schema validation at all**. It reads `await request.json()` into an `any`, destructures whatever it finds, and hands the raw values straight to `prisma.user.update()`.

https://github.com/TravellersMeet/travellers/blob/main/src/app/api/user/onboard/route.ts

```ts
let body: any = {};
try {
  if (request && typeof request.json === "function") {
    body = await request.json();
  }
} catch (e) {
  // ignore
}

const { name, languages, travelInterests, /* ... */ } = body;

await prisma.user.update({
  where: { id: session.user.id },
  data: {
    onboarded: true,
    name: name ?? undefined,
    languages: languages ?? undefined,
    ...
  },
});
```

Every other profile-writing path (`PATCH /api/user/profile`) at minimum type-checks each field before it reaches Prisma. Onboarding does not, and it is the endpoint every new account goes through.

## Problems

1. **No type checking.** `languages`, `travelInterests`, `accommodationPrefs` and `socialLinks` are `String[]` columns. Posting `{"languages": [1, 2, 3]}` or `{"languages": "hindi"}` sends a malformed value into Prisma, which throws `PrismaClientValidationError` and surfaces as an opaque `500 Failed to complete onboarding`.
2. **No length bounds.** `bio`, `location`, `homeLocation`, `budgetRange`, `gender`, `travelStyle` and every array element are unbounded. `bio` is `@db.Text`, so a single request can persist megabytes, and the arrays have no element cap — `{"travelInterests": [ ...50000 items ]}` is accepted.
3. **`name` can be emptied.** `name` is a non-nullable column but `""` passes straight through, leaving accounts with a blank display name that then renders as an empty card everywhere in the match list and chat sidebar.
4. **`age` is barely parsed.** `parseInt(age.toString(), 10)` accepts `"42abc"` → `42`, and accepts `-5` or `99999` unchanged. There is no plausible-range check.
5. **`socialLinks` is not validated as URLs.** Arbitrary strings — including `javascript:` payloads — are stored and later rendered as profile links.
6. **Soft-deleted accounts can re-onboard.** `PATCH /api/user/profile` refuses to act when `isDeleted` is true. `POST /api/user/onboard` never checks, so a still-valid session belonging to a deleted account can write to the row and flip `onboarded` back to true.
7. **No rate limit.** Every other mutating route in the codebase goes through `enforceRateLimit`. This one can be looped freely, and each call is an unbounded write.
8. **Swallowed parse errors.** A malformed body is silently treated as `{}`, so the user is told onboarding succeeded while nothing but `onboarded: true` was written.

## Steps to reproduce

```bash
# 1. Blank out the display name
curl -X POST /api/user/onboard -H 'content-type: application/json' \
  -b "$SESSION" -d '{"name":""}'
# -> 200 {"success":true}; users.name is now ""

# 2. Malformed array -> opaque 500
curl -X POST /api/user/onboard -H 'content-type: application/json' \
  -b "$SESSION" -d '{"languages":"hindi"}'
# -> 500 {"error":"Failed to complete onboarding"}

# 3. Implausible age accepted
curl -X POST /api/user/onboard -H 'content-type: application/json' \
  -b "$SESSION" -d '{"age":"42-not-a-number"}'
# -> 200; users.age is now 42
```

## Expected behaviour

- A Zod schema validates the body, with explicit bounds on every field and a `400` carrying field-level details when it fails.
- `name` must be non-empty after trimming.
- `age` is an integer inside a plausible range, rejected rather than silently coerced.
- `socialLinks` entries must be `http(s)` URLs.
- Arrays have an element count cap and a per-element length cap.
- Soft-deleted accounts get `404`, matching `PATCH /api/user/profile`.
- The route is rate limited like every other mutating endpoint.
- A malformed JSON body returns `400`, not a fake success.

## Suggested approach

Extract the shared field bounds into `src/lib/validation/profile.ts` so onboarding and the profile PATCH can converge on one source of truth, then run the route through the existing `withValidation` wrapper and add a `userOnboard` rule to `RATE_LIMIT_RULES`.

MOHITKOURAV01:
Closes #423

## What was wrong

`POST /api/user/onboard` was the only user-write endpoint with no schema at all. It read the body into `any`, destructured it, and handed the raw values to `prisma.user.update()`:

```ts
let body: any = {};
try { body = await request.json(); } catch (e) { /* ignore */ }
const { name, languages, age, socialLinks, ... } = body;
await prisma.user.update({ where: { id: session.user.id }, data: { onboarded: true, name: name ?? undefined, ... } });
```

Concretely that allowed:

| Payload | Old behaviour |
| --- | --- |
| `{"name": ""}` | 200, `users.name` blanked — renders as an empty card in matches and chat |
| `{"languages": "hindi"}` | `PrismaClientValidationError` → opaque `500` |
| `{"age": "42-not-a-number"}` | 200, age stored as `42` |
| `{"bio": "<5 MB>"}` | 200, persisted to the `@db.Text` column |
| `{"socialLinks": ["javascript:alert(1)"]}` | 200, stored and later rendered as an anchor |
| `{...}` from a soft-deleted account | 200, row written and `onboarded` flipped back on |
| malformed JSON | 200 `{"success": true}` with nothing but `onboarded: true` written |

## What this changes

**`src/lib/validation/profile.ts` (new)** — the field bounds for every endpoint that writes the user profile, in one place so onboarding and `PATCH /api/user/profile` can't drift:

- `name` — trimmed, whitespace-collapsed, 2–80 chars, cannot end up empty
- `bio` / `location` / `homeLocation` / `budgetRange` / `gender` / `travelStyle` — trimmed with explicit maximums
- `languages` / `travelInterests` / `accommodationPrefs` — arrays of short tags, capped at 25 entries × 60 chars; blank entries the form leaves behind are dropped rather than rejected
- `socialLinks` — must parse as a URL with an `http(s)` protocol, max 10
- `age` — accepts a number or a fully numeric string, rejects `"42abc"` and fractions, bounded to 13–120
- `.strict()` so an unknown key is a visible 400 rather than a silent no-op — this also means the payload can no longer smuggle in `role` or `onboarded`

**`src/app/api/user/onboard/route.ts`** — runs through the existing `withValidation` wrapper, so malformed JSON and schema failures both return `400` with field-level details. It also now:

- looks the account up first and returns `404` when `isDeleted` is true, matching `PATCH /api/user/profile`
- rate limits through a new `userOnboard` rule (10 per 10 min, keyed on user + IP), in line with every other mutating route
- writes only the keys the client actually sent, instead of passing `undefined` for the rest
- returns the updated profile so the client doesn't need a follow-up `GET`

**`src/lib/rate-limit-rules.ts`** — adds the `userOnboard` entry.

## Tests

45 new cases across two files:

- `src/lib/__tests__/profile-validation.test.ts` (26) — every bound, both sides of each boundary, the `javascript:`/`data:` link rejection, the `"42abc"` coercion refusal, and unknown-key rejection.
- `src/app/api/user/onboard/__tests__/route.test.ts` (19) — auth, the happy path, partial payloads, each validation failure returning 400 without touching the database, the soft-delete 404, and the rate-limit 429 with `Retry-After`.

Two existing tests asserted the old behaviour and are updated:

- `profile-expansion.test.ts` — now mocks the soft-delete lookup and relaxes the `update` assertion to `objectContaining`, since the handler passes a `select` alongside `data`.
- `moderation.test.ts` — passes a real request shape instead of `{}` (the route now reads headers for rate limiting), and gains a case covering the soft-deleted 404.

## Verification

```
$ npx vitest run
Test Files  66 passed (66)
     Tests  514 passed (514)

$ npx tsc --noEmit
# no new errors — the 6 remaining are pre-existing on main and come from a
# stale generated Prisma client (verificationStatus / metadataHash),
# unrelated to this change
```

## Notes for review

- The onboarding modal (`src/components/OnboardingModal.tsx`) already sends exactly this set of keys, so `.strict()` does not break the current client. It builds `socialLinks` as full `https://instagram.com/...` / `https://x.com/...` URLs, which pass the new protocol check.
- `toProfileUpdateData` is typed against `Record<string, unknown>` rather than the inferred schema type: `age` goes through a transform, so the wrapper's inferred payload type keeps the pre-transform union and would not line up with the output type.
- The bounds are deliberately generous. The goal is to make a runaway or malicious payload impossible, not to argue about whether a bio should be 1 000 or 2 000 characters — happy to adjust any of the numbers in `PROFILE_LIMITS`.
- `PATCH /api/user/profile` still has its own hand-rolled type checks. Pointing it at the same schema is the obvious follow-up, but it changes a second endpoint's error shape, so I kept it out of this PR.


santichausis (approved):
Read through the validation module and the route handler in full, not just the description — this is solid.

- The schema bounds are sensible and well-justified (the `.finite()`/numeric-string rejection on age, the `javascript:`/`data:` protocol check on social links, `.strict()` to surface client drift instead of silently dropping unknown keys).
- Soft-delete guard matches the existing pattern in `PATCH /api/user/profile`.
- The 45 new tests are genuinely exercising each branch (every validation bound both sides, soft-delete, rate limit, DB failure) rather than being tautological.
- CI is green (`build-and-lint: pass`).

Nothing to flag. Nice work.

singh-odyssey (approved):

PR #429, opened 2026-08-26
fix(heatmap): geocode ticket destinations instead of hashing string length

### LINKED ISSUE #424: [Bug] GET /api/matches/heatmap fabricates coordinates from string length instead of geocoding destinations
## Summary

`GET /api/matches/heatmap` does not geocode anything. It generates coordinates from the **character length of the destination string** and the array index, then serves the result to `TravelHeatmap` as if it were real traveller data.

https://github.com/TravellersMeet/travellers/blob/main/src/app/api/matches/heatmap/route.ts

```ts
const features = tickets.map((t, i) => {
  // Very simple determinisitic pseudo-random coordinate generation for demo purposes
  // based on the string length to give varying hotspots
  const hash = t.destination.length + i;
  const lat = 35 + (hash % 15) * (hash % 2 === 0 ? 1 : -1);
  const lng = -10 + (hash % 25) * (hash % 3 === 0 ? 1 : -1);
  ...
});
```

The component renders this under the label **"🔥 Global Traveller Hotspots"** (`src/components/spatial/TravelHeatmap.tsx`), so the dashboard is presenting fabricated geography as a real product signal.

## Why this is wrong

1. **The output is meaningless.** `"Goa"` (4 chars) and `"Bali"` (4 chars) land on the exact same point; `"Tokyo"` and `"Delhi"` (5 chars each) also collide. Two genuinely different hotspots become one, and one destination splits across many points because the index `i` is mixed into the hash — the same city moves every time the ticket ordering changes.
2. **The coordinates aren't even on land.** `lat` is confined to 20–50 and `lng` to −35…+15, so every "global" hotspot falls in a box over the North Atlantic, western Europe and north Africa. A dashboard full of Asian and South American tickets renders a heat blob off the coast of Portugal.
3. **Nothing is aggregated.** `weight` is hardcoded to `1` per ticket and the layer paint also hardcodes `"heatmap-weight": 1`, so a destination with 400 verified tickets is indistinguishable from one with a single ticket. The one thing a heatmap exists to show — relative density — is discarded.
4. **The query is unbounded.** `prisma.ticket.findMany` selects *every* verified future ticket in the system with no `take`, and one GeoJSON feature is emitted per ticket. This grows linearly with platform size and is recomputed on every dashboard load with no caching.
5. **The result is unstable across requests.** Because `i` is part of the hash, the same database state produces different coordinates depending on row ordering.

The code comments acknowledge this is placeholder work ("for demo purposes", "for the sake of the WOW factor PR"), but it is live on the dashboard.

## The data to do it properly already exists

`src/lib/data/destinations.ts` is a 500-line gazetteer of major destinations with real coordinates, countries, regions and alias lists:

```ts
{
  id: 'nyc',
  name: 'New York City, NY, USA',
  city: 'New York City',
  coordinates: { lat: 40.7128, lng: -74.0060 },
  aliases: ['NYC', 'New York', 'Manhattan'],
}
```

and `src/lib/normalize-destination.ts` already provides the normalisation used elsewhere for ticket destination matching.

## Expected behaviour

- Ticket destinations are resolved against the existing destinations dataset (name, city, aliases, airport code), using `normalizeDestination` so `"goa "`, `"GOA"` and `"Goa"` collapse to one entry.
- Tickets for the same destination are **aggregated into one feature** whose `weight` is the ticket count, and the MapLibre layer reads that property instead of a hardcoded `1`.
- Destinations that cannot be resolved are excluded from the map and reported as a count, rather than being placed at a fake coordinate.
- The query is bounded and the response is cached for a short TTL, in line with how `lib/match-cache.ts` treats other aggregate reads.

## Steps to reproduce

1. Create verified tickets for `Tokyo` and `Delhi` with future departure dates.
2. `GET /api/matches/heatmap`.
3. Both render at the same latitude/longitude band in the Atlantic, and neither is anywhere near Japan or India.

MOHITKOURAV01:
Closes #424

## What was wrong

`GET /api/matches/heatmap` never geocoded anything. It derived coordinates from the **character length of the destination string** plus the array index:

```ts
const hash = t.destination.length + i;
const lat = 35 + (hash % 15) * (hash % 2 === 0 ? 1 : -1);
const lng = -10 + (hash % 25) * (hash % 3 === 0 ? 1 : -1);
```

The dashboard renders that under "🔥 Global Traveller Hotspots".

Four separate consequences:

1. **Collisions.** `"Goa"`/`"Bali"` (4 chars) and `"Tokyo"`/`"Delhi"` (5 chars) resolved to the same point.
2. **Wrong hemisphere.** `lat` was confined to 20–50 and `lng` to −35…+15, so every "global" hotspot fell in a box over the North Atlantic and western Europe. Asian and South American tickets rendered off the coast of Portugal.
3. **No density.** `weight` was hardcoded to `1` per ticket, and the layer paint also hardcoded `"heatmap-weight": 1` — a destination with 400 tickets looked identical to one with 1.
4. **Unstable.** The row index `i` was part of the hash, so identical data moved between requests.

The query also had no `take`, emitting one feature per verified ticket platform-wide on every dashboard load.

### `src/lib/geocode-destination.ts` (new)

The repo already ships a 500-line gazetteer at `src/lib/data/destinations.ts` with real coordinates, countries and alias lists. This module indexes it once and resolves the free-text `Ticket.destination` against it:

- Index keys: id, full name, city, `"City, Country"`, every alias, and the IATA airport code.
- Normalisation goes through the existing `normalizeDestination`, so `"tokyo"`, `"TOKYO"` and `"  Tokyo  "` collapse to one entry.
- Comma-separated inputs are tried longest-prefix first, then segment by segment, so both `"Tokyo, Japan"` and `"Shinjuku, Tokyo, Japan"` resolve without falling back to substring matching (which would let `"Osaka"` match inside an unrelated string).
- **A key claimed by two different cities is dropped, not guessed** — silently attributing tickets to the wrong hotspot is the failure mode this PR exists to remove.
- **Anything the gazetteer doesn't know returns `null`.** It is not placed at an approximate coordinate.

`aggregateDestinations()` groups raw strings into one bucket per real place with a count, sorted heaviest-first with a deterministic tiebreak.

### `src/app/api/matches/heatmap/route.ts`

- One GeoJSON feature per destination, `properties.weight` = ticket count, plus id/name/city/country/region for labelling.
- `take: 5000` on the query, ordered by `departureDate asc` so the cap trims the far future rather than imminent trips.
- Features capped at 250, heaviest kept.
- A `meta` block reporting `sampledTickets`, `resolvedTickets`, `unresolvedTickets`, `unresolvedDestinations`, `maxWeight` and `truncated` — gazetteer gaps stay visible instead of vanishing.

### `src/components/spatial/TravelHeatmap.tsx`

`"heatmap-weight"` now interpolates on `["get", "weight"]` against `meta.maxWeight` instead of the hardcoded `1`, so one busy city doesn't flatten the ramp. The `any` state is replaced with a typed `HeatmapCollection`.

## Tests

39 new cases:

- `src/lib/__tests__/geocode-destination.test.ts` (22) — exact/alias/airport/`City, Country` resolution, case and whitespace handling, the substring-match rejection, `null` for unknown and non-string input, aggregation counting and ordering, unresolved reporting, plus a gazetteer integrity check that every coordinate is within valid bounds and every id is unique.
- `src/app/api/matches/heatmap/__tests__/route.test.ts` (17) — auth, real coordinates (Tokyo asserted at 139.65/35.68), the explicit **"Tokyo and Delhi no longer collide"** regression, order-independence, weight aggregation across `"Tokyo"`/`"tokyo"`/`"Tokyo, Japan"`/`"NRT"`, unresolved destinations excluded but reported, and the query bounds.

One test caught a real bug during development: the unresolved-label map kept the *last* spelling seen, so the reported label changed with row ordering. It now keeps the first.

## Verification

```
$ npx vitest run
Test Files  65 passed (65)
     Tests  462 passed (462)

$ npx tsc --noEmit
# client: verificationStatus / metadataHash), unrelated to this change
```

## Notes for review

- **The gazetteer has 40 entries, so real destinations will go unresolved** — "Goa" isn't in it today. That's deliberate: the endpoint now under-reports rather than lying about where people are going, and `meta.unresolvedDestinations` makes the gap measurable. Expanding the dataset is a natural follow-up, and the shape of it means new entries need no code change.
- Uploads have no destination autocomplete enforcement, so free text will keep arriving. `resolveDestination` is written to be reused by the ticket upload path if we later want to nudge users toward a known destination at entry time.
- I left caching out. The read is now bounded and cheap, and `lib/match-cache.ts` is keyed per-user, which doesn't fit a global aggregate. If the query shows up in profiling, a short flat-key TTL is easy to add.
- `HEATMAP_TICKET_LIMIT` (5 000) and `HEATMAP_MAX_FEATURES` (250) are exported so the tests assert against them rather than duplicating the numbers — both are guesses at a sane ceiling and easy to change.


MOHITKOURAV01:
CI caught something I'd missed locally: a Next.js App Router route file may only export route handlers and the framework's own config fields, so exporting `HEATMAP_TICKET_LIMIT` and `HEATMAP_MAX_FEATURES` from `route.ts` failed the build.

```
Type error: Route "src/app/api/matches/heatmap/route.ts" does not match the required types of a Next.js Route.
  "HEATMAP_TICKET_LIMIT" is not a valid Route export field.
```

`tsc --noEmit` doesn't flag this — only `next build` does, which is why it got past my local check.

Moved both constants into `./constants.ts`, matching the existing convention in `src/app/api/meetup-checklist/constants.ts`, so the tests still assert against them rather than duplicating the numbers. Build is green now.

PR #430, opened 2026-08-26
fix(routes): bound the saved-route payload and cap route writes

### LINKED ISSUE #425: [Bug] POST /api/routes accepts out-of-range coordinates, non-finite distances and unbounded payloads
## Summary

The Zod schema behind `POST /api/routes` checks *types* but never checks *values*. Latitude and longitude are `z.number()` with no range, `distance` and `duration` are `z.number()` so `NaN`/`Infinity`/negatives pass, and every string and array in the payload is unbounded.

https://github.com/TravellersMeet/travellers/blob/main/src/app/api/routes/route.ts

```ts
const RouteSchema = z.object({
  origin: z.object({ lat: z.number(), lng: z.number() }),
  destination: z.object({ lat: z.number(), lng: z.number() }),
  waypoints: z.array(z.object({
    location: z.object({ lat: z.number(), lng: z.number() }),
    stopover: z.boolean(),
    name: z.string().optional(),
  })).optional(),
  originName: z.string().optional(),
  destinationName: z.string().optional(),
  distance: z.number(),
  duration: z.number(),
  encodedPolyline: z.string(),
  tripName: z.string().optional(),
  notes: z.string().optional(),
});
```

## Problems

1. **Coordinates are unbounded.** `{"lat": 900, "lng": -4000}` is stored. `Route.originLat` etc. are plain `Float` columns with no DB constraint, so the bad value survives to the map renderer, where MapLibre either clamps it silently or drops the layer.
2. **`NaN` and `Infinity` are accepted.** `z.number()` passes both. `distance: Infinity` and `duration: NaN` reach Prisma, and `NaN` then serialises to `null` in the JSON response — so a route reads back differently from what was written.
3. **Negative distance/duration are accepted.** `distance: -1` is not physically meaningful and is used directly by `TripBudgetEstimator` and the shared-route card in chat.
4. **`encodedPolyline` is unbounded and `@db.Text`.** Nothing stops a client from posting a 50 MB string. There is no size ceiling anywhere between the request body and the column.
5. **`waypoints` has no element cap** and is `JSON.stringify`-ed straight into another `@db.Text` column. A 100 000-element array is accepted in one request.
6. **`tripName`, `notes`, `originName`, `destinationName` have no max length**, while `tripName`/`originName`/`destinationName` map to plain `String` columns (varchar) — an oversized value is a database error surfaced as a generic `500 Failed to save route`.
7. **No per-user route cap and no rate limit.** `POST /api/routes` is the only remaining mutating route with neither. `messageSend`, `connectionAction`, `userReport` and the auth routes all go through `enforceRateLimit`; this one can be looped to fill the table.
8. **`stopover` is required on waypoints** but the client (`maplibre-route.tsx`) does not always send it, so valid-looking payloads are rejected with a confusing `Invalid input`.

## Steps to reproduce

```bash
# Impossible coordinates are persisted
curl -X POST /api/routes -H 'content-type: application/json' -b "$SESSION" -d '{
  "origin": {"lat": 900, "lng": -4000},
  "destination": {"lat": -900, "lng": 4000},
  "distance": -1, "duration": 0,
  "encodedPolyline": "abc"
}'
# -> 201, row written with lat 900

# NaN round-trips as null
curl -X POST /api/routes ... -d '{..., "distance": 1e400, "duration": 5}'
# -> 201; GET /api/routes returns "distance": null
```

## Expected behaviour

- `lat` ∈ [−90, 90], `lng` ∈ [−180, 180], enforced for origin, destination and every waypoint.
- `distance` and `duration` must be finite, non-negative, and below a sane ceiling.
- Explicit max lengths on `encodedPolyline`, `tripName`, `notes`, `originName`, `destinationName` and waypoint names.
- A cap on the number of waypoints per route.
- A per-user cap on stored routes, returning `409` with a clear message rather than growing without bound.
- `stopover` defaults to `true` instead of being required.
- The route is rate limited through the existing `RATE_LIMIT_RULES` table.
- Validation failures return `400` with field-level details, not a generic `500`.

MOHITKOURAV01:
Closes #425

## What was wrong

`RouteSchema` checked *types* but never *values*:

```ts
origin: z.object({ lat: z.number(), lng: z.number() }),
distance: z.number(),
duration: z.number(),
encodedPolyline: z.string(),
notes: z.string().optional(),
```

`Route.originLat` and friends are plain `Float` columns with no database constraint, and `encodedPolyline` / `notes` / `waypoints` are `@db.Text`. The request schema was the only thing standing between a client and those columns, and it wasn't checking anything.

| Payload | Old behaviour |
| --- | --- |
| `{"lat": 900, "lng": -4000}` | `201`, stored; MapLibre later drops or clamps the layer |
| `"distance": -1` | `201`, and feeds `TripBudgetEstimator` |
| `"distance": 1e400` | `201`, `Infinity`/`NaN` written — then **reads back as `null`** |
| 50 MB `encodedPolyline` | `201` |
| 100 000 waypoints | `201`, `JSON.stringify`-ed into a Text column |
| oversized `tripName` | database error → generic `500 Failed to save route` |
| waypoint without `stopover` | `400 Invalid input` on an otherwise valid payload |

There was also no rate limit and no per-user ceiling — the only remaining mutating route with neither.

### `src/lib/validation/route-payload.ts` (new)

| Field | Bound |
| --- | --- |
| `lat` | finite, −90…90 |
| `lng` | finite, −180…180 |
| `distance` | finite, 0…40 075 000 m (equatorial circumference) |
| `duration` | finite, 0…31 536 000 s (one year) |
| `encodedPolyline` | 1…200 000 chars |
| `waypoints` | ≤ 25, each coordinate validated |
| `tripName` | ≤ 120 |
| `notes` | ≤ 2 000 |
| `originName` / `destinationName` / waypoint `name` | ≤ 200 |

`.finite()` is the load-bearing part: `z.number()` accepts both `NaN` and `Infinity`, which is how a route ended up reading back differently from how it was written.

`stopover` now **defaults to `true`** rather than being required — `src/lib/types/route.ts` declares it required but the map client omits it for waypoints dragged onto the line, so valid payloads were being rejected.

### `src/app/api/routes/route.ts`

- Uses the shared schema, so bad input is a `400` with field-level details instead of a `500`.
- Adds a `routeWrite` rate limit (30/min, keyed on user + IP).
- Adds a 200-route per-user ceiling on **creates**, returning `409` with an actionable message. Updates are exempt — they replace a row rather than add one, so editing still works at the ceiling.
- The create and update paths now share one `routeData` object, so a route reads back identically on either. Previously the two field lists were duplicated and could drift.
- The ownership lookup selects `{ id: true }` instead of the whole row.

## Tests

49 new cases:

- `src/lib/__tests__/route-payload-validation.test.ts` (31) — every bound with both sides of each boundary, `NaN`/`±Infinity` rejection, numeric strings not being coerced, the `stopover` default, and a check that the error path points at `origin.lat` so the client can highlight the field.
- `src/app/api/routes/__tests__/create-route.test.ts` (18) — auth, create, update, the cross-user `404`, each bound rejected *without touching the database*, the `409` at the ceiling, updates still allowed at the ceiling, and the rate-limit `429` with `Retry-After`.

The `1e400` case is written as a raw JSON body rather than a JS `NaN` literal, since that's how the value actually arrives — JSON has no `NaN`, but `1e400` parses to `Infinity`.

## Verification

```
$ npx vitest run
Test Files  65 passed (65)
     Tests  472 passed (472)

$ npx tsc --noEmit
# client: verificationStatus / metadataHash), unrelated to this change
```

The existing `src/app/api/routes/__tests__/pagination.test.ts` still passes untouched.

## Notes for review

- **The ceilings are judgement calls.** 200 routes/user, 25 waypoints, 200 KB polyline — all sized to be far above any real trip while making a runaway payload impossible. They're constants in `ROUTE_LIMITS` and the tests derive from them, so changing a number needs no test edits.
- **`routesPerUser` costs one `count()` per create.** `Route` is indexed on `[userId, updatedAt]`, so it's cheap, but it is an extra round-trip. I judged that worth it against an uncapped table; happy to drop it if you'd rather not pay for it on every save.
- Existing rows written before this change can still hold out-of-range values. This only closes the door — a backfill to find and flag them is a separate piece of work, and I didn't want to put a data migration in a validation PR.
- `DELETE /api/routes` is untouched.


santichausis (approved):
Also read through this one in full — same bar as #428.

- Good catch on `NaN`/`Infinity` reaching Prisma via a bare `z.number()` and reading back as `null` — the `.finite()` bound plus the `1e400`-as-JSON test case for it is exactly the right way to pin that down.
- The create/update paths sharing one `routeData` object so they can't drift is a nice cleanup on top of the validation fix.
- Per-user ceiling correctly exempts updates (verified the "still allows an update at the ceiling" test matches the route.ts logic), and the ownership check on update (404 for another user's route) is intact.
- Tests are thorough and non-trivial — bounds, ownership, ceiling, rate limiting, DB failure.
- CI is green (`build-and-lint: pass`).

Nothing to flag here either. Approved.

singh-odyssey (approved):

PR #432, opened 2026-08-26
fix(conversations): paginate the sidebar list and stabilise its ordering

### LINKED ISSUE #427: [Bug] GET /api/conversations loads every conversation with no pagination or stable ordering
## Summary

`GET /api/conversations` loads **every** conversation the user belongs to in a single unbounded query, and for each one eagerly includes the participants and the latest message. It is now the last list endpoint in the API without pagination.

https://github.com/TravellersMeet/travellers/blob/main/src/app/api/conversations/route.ts

```ts
const conversations = await prisma.conversation.findMany({
  where: { users: { some: { id: userId }, ... } },
  include: {
    users: { where: { id: { not: userId } }, select: { ... } },
    messages: { orderBy: { createdAt: "desc" }, take: 1 },
  },
  orderBy: { updatedAt: "desc" },
});
```

There is no `take`, no `cursor`, and no pagination metadata in the response.

## Why this matters now

The rest of the API has already standardised on cursor pagination through `src/lib/pagination.ts`:

| Endpoint | Paginated |
| --- | --- |
| `GET /api/messages` | yes (#350) |
| `GET /api/notifications` | yes (#381) |
| `GET /api/routes` | yes |
| `GET /api/conversations` | **no** |

The chat sidebar (`src/components/chat/`) calls this endpoint on every mount. An active user with a few hundred threads pulls the entire list plus a joined message row per thread on every page load, and the payload only grows — conversations are never archived or deleted anywhere in the codebase.

## Additional problems in the same handler

1. **No `id` tiebreaker in the ordering.** `orderBy: { updatedAt: "desc" }` alone is not a stable sort. Conversations created in the same transaction share a timestamp and can swap places between requests — which is also exactly why a cursor cannot be built on `updatedAt` alone. `lib/pagination.ts` already solves this with its `(timestamp, id)` composite cursor.
2. **`otherUser` silently becomes `null` for self-conversations.** `conv.users[0] || null` — the `users` include filters out the caller, so any conversation where the caller is the only participant yields `otherUser: null`, and the sidebar renders a blank row instead of skipping it.
3. **The blocked-user filter runs on every request with no reuse.** `getBlockedUserIds(userId)` is called before the query on each load; with pagination in place it is worth confirming the filter still applies per page rather than only to the first.

## Steps to reproduce

1. Seed ~300 conversations for one user.
2. Open the chat sidebar and watch `GET /api/conversations`.
3. The response contains all 300 entries with no `pagination` block, and there is no way for the client to ask for fewer.

## Expected behaviour

- The endpoint accepts `limit` and `cursor` via `parsePaginationParams`, capped by the existing `MAX_PAGE_LIMIT`.
- Ordering is `[{ updatedAt: "desc" }, { id: "desc" }]` so the cursor is stable.
- The response includes the standard `pagination: { limit, nextCursor, hasMore }` block, while keeping the existing `conversations` key so the current client keeps working.
- An invalid cursor or limit returns `400` via `PaginationError`, consistent with `/api/messages` and `/api/routes`.
- Conversations with no counterpart are skipped rather than emitted with `otherUser: null`.

MOHITKOURAV01:
Closes #427

## What was wrong

`GET /api/conversations` loaded **every** conversation the caller belonged to in one unbounded query, eagerly including the participants and the latest message for each:

```ts
const conversations = await prisma.conversation.findMany({
  where: { users: { some: { id: userId }, ... } },
  include: {
    users: { where: { id: { not: userId } }, select: { ... } },
    messages: { orderBy: { createdAt: "desc" }, take: 1 },
  },
  orderBy: { updatedAt: "desc" },
});
```

No `take`, no `cursor`, no pagination metadata. `ChatInterface` calls it on every mount, and conversations are never archived or deleted anywhere in the codebase, so the payload only grows.

It was the last list endpoint without pagination:

| Endpoint | Before | After |
| --- | --- | --- |
| `GET /api/messages` | paginated (#350) | — |
| `GET /api/notifications` | paginated (#381) | — |
| `GET /api/routes` | paginated | — |
| `GET /api/conversations` | **unbounded** | paginated |

### Cursor pagination

Uses the existing `src/lib/pagination.ts` — `parsePaginationParams`, `buildTimestampCursorWhere`, `createPaginatedResponse` — so the response shape matches the other three endpoints exactly. An invalid cursor or limit returns `400` via `PaginationError`.

### Stable ordering

`orderBy` becomes `[{ updatedAt: "desc" }, { id: "desc" }]`. The `id` tiebreaker matters twice over: `updatedAt` alone is not a stable sort for conversations created in the same transaction, and **without it the cursor could skip or repeat rows** across pages.

### Filtering order

Pagination is applied to the raw query result *before* the payload is shaped:

```ts
const page = createPaginatedResponse(conversations, limit, "updatedAt");
const formatted = page.items.filter(...).map(...);
```

Filtering first would corrupt both `hasMore` and `nextCursor` — they need to describe the query, not the trimmed list. There's a test pinning this: a page where one row is dropped still reports `hasMore: true` and a cursor pointing at the last *queried* row.

### Counterpart handling

A conversation where the caller is the only participant was emitted as `otherUser: null`. `ChatInterface` already did `if (!other) return null`, so it rendered nothing — but it still consumed a page slot and crossed the wire. It's now dropped server-side.

### Block filter

Unchanged in behaviour, but explicitly covered on a cursored page — the `none` clause is part of the query `where`, so it holds on every page rather than only the first. That was the open question in the issue.

### Client

`ChatInterface` stores `pagination.nextCursor` and gains a **"Load older conversations"** button at the foot of the sidebar, mirroring how older messages are already paged in. It de-duplicates on append, since a thread can move between pages if it receives a message mid-scroll.

The `conversations` response key is unchanged, so nothing else needs touching.

## Tests

15 new cases in `src/app/api/conversations/__tests__/pagination.test.ts`:

- Default limit, explicit limit, clamping above `MAX_PAGE_LIMIT`
- The `[updatedAt, id]` ordering
- `hasMore: false` on a short page; `hasMore: true` plus a cursor on a full one
- The cursor decoding to the last item of the page
- A supplied cursor producing the right `OR` clause
- The block filter surviving alongside a cursor
- `400` for an invalid cursor, a non-numeric limit, and `limit=0`
- The `conversations` key keeping its existing shape
- A skipped conversation counting against the page but not the cursor
- `500` on a query failure

One existing test in `route.test.ts` asserted `otherUser: null` for a conversation with no participants. That's the behaviour this PR removes, so it's split into two: one confirming `lastMessage: null` still works for a real conversation with no messages, and one confirming the counterpart-less conversation is now skipped entirely.

## Verification

```
$ npx vitest run
Test Files  64 passed (64)
     Tests  439 passed (439)

$ npx tsc --noEmit
# client), unrelated to this change

$ npx next lint --file src/components/ChatInterface.tsx
✔ No ESLint warnings or errors
```

## Notes for review

- **This changes the default response for existing clients**: they now get 20 conversations instead of all of them. `ChatInterface` is the only caller in the repo and it's updated here, but it's worth a second pair of eyes in case anything external reads this endpoint.
- The default limit is `DEFAULT_PAGE_LIMIT` (20) from the shared module rather than a conversation-specific number, on the grounds that consistency with the other three endpoints beats tuning this one in isolation.
- A "load older" button is the least surprising option and matches the message history. Infinite scroll on the sidebar would be nicer, but it needs an intersection observer and scroll-anchoring work that doesn't belong in this PR.
- The `COUNTERPART_SELECT` extraction is cosmetic — same columns as before, just named so the intent of keeping it narrow is visible.


santichausis (approved):
Read the route and the ChatInterface changes in full. Correct — the shared `pagination.ts` helpers are used exactly as in the other three endpoints, the `[updatedAt, id]` tiebreaker is the right fix for the stability problem, and the counterpart filter happening *after* `createPaginatedResponse` (not before) is exactly what keeps `hasMore`/`nextCursor` honest. The client-side de-dup on append is a sensible guard for a thread moving between pages.

singh-odyssey (approved):

PR #438, opened 2026-08-28
fix(rate-limit): make the documented RATE_LIMIT_* environment variables actually apply

### LINKED ISSUE #433: [Bug] Documented RATE_LIMIT_AUTH_* environment variables have no effect — the env-driven config table is dead code
## Summary

`.env.example` documents ten `RATE_LIMIT_AUTH_*` knobs, and `src/lib/rate-limit-config.ts` reads all of them. None of them change the limits that are actually enforced. Setting `RATE_LIMIT_AUTH_SIGNIN_LIMIT=3` in production has no effect whatsoever — sign-in stays at 10 attempts per 10 minutes.

## Why

There are two rate-limit tables in the repo and only one of them is wired up.

`src/lib/rate-limit-config.ts` builds `RATE_LIMIT_CONFIG` from environment variables:

```ts
export const RATE_LIMIT_CONFIG = {
  auth: {
    signin: {
      limit: getEnvNumber("RATE_LIMIT_AUTH_SIGNIN_LIMIT", 10),
      windowSeconds: getEnvNumber("RATE_LIMIT_AUTH_SIGNIN_WINDOW_SECONDS", 600),
    },
    ...
```

`src/lib/rate-limit-rules.ts` builds `RATE_LIMIT_RULES` from hardcoded literals:

```ts
export const RATE_LIMIT_RULES = {
  authSignin: {
    namespace: "auth:signin",
    limit: 10,           // <- not read from the environment
    windowSeconds: 10 * 60,
  },
  ...
```

Every route goes through `enforceRateLimit()`, which reads `RATE_LIMIT_RULES`. Grep for the other one:

```
$ grep -rn "RATE_LIMIT_CONFIG" src
src/app/api/auth/__tests__/signin.test.ts:25 ...
src/lib/__tests__/rate-limit-config.test.ts:2 ...
src/lib/rate-limit-config.ts:7 ...
```

The only consumers are its own tests. It is dead code that reads live environment variables, which is the worst combination — it looks configured and tests green while doing nothing.

## Impact

- Operators cannot tighten limits during an attack without a code change and redeploy.
- The eight rules that only exist in `RATE_LIMIT_RULES` (`authResetPassword`, `authChangePassword`, `chat`, `messageSend`, `connectionAction`, `userReport`, …) have no environment override at all — including `chat`, which fronts a per-request billed Gemini call.
- `src/lib/__tests__/rate-limit-config.test.ts` asserts the defaults of a table nothing enforces, so the suite reports confidence it has not earned.
- `getEnvNumber` silently swallows bad input: `RATE_LIMIT_AUTH_SIGNIN_LIMIT=abc` falls back to the default with no warning, so a typo in a deploy config is invisible.

## Steps to reproduce

1. `RATE_LIMIT_AUTH_SIGNIN_LIMIT=2 npm run dev`
2. `POST /api/auth/signin` with bad credentials 3 times from one IP.
3. Expected: 429 on the 3rd. Actual: 200/401 through to the 11th.

## Proposed fix

Make `RATE_LIMIT_RULES` the single env-driven source of truth:

- Give every rule an env key pair (`RATE_LIMIT_<RULE>_LIMIT` / `_WINDOW_SECONDS`), derived from the rule name so the table and the keys cannot drift.
- Warn once on an unparseable value instead of silently falling back.
- Keep `RATE_LIMIT_CONFIG` exported as a thin view over the resolved rules so existing imports and tests keep working, and mark it deprecated.
- Extend `.env.example` to cover every rule, not just the five auth ones.
- Add tests that resolve the table under a mocked `process.env` and prove an override actually lands.

MOHITKOURAV01:
Closes #433

## The problem

Two rate-limit tables, one enforced.

Every throttled route calls `enforceRateLimit()`, which reads `RATE_LIMIT_RULES` — hardcoded literals. `src/lib/rate-limit-config.ts` built a *separate* table from `RATE_LIMIT_AUTH_*` environment variables and nothing outside its own tests ever imported it:

```
$ grep -rn "RATE_LIMIT_CONFIG" src
src/app/api/auth/__tests__/signin.test.ts:25 ...
src/lib/__tests__/rate-limit-config.test.ts:2 ...
src/lib/rate-limit-config.ts:7 ...
```

So the ten knobs documented in `.env.example` parsed correctly, tested green, and did nothing. `RATE_LIMIT_AUTH_SIGNIN_LIMIT=2` still let 10 attempts through.

## The change

`RATE_LIMIT_RULES` becomes the single source of truth and resolves overrides at module load.

**`RATE_LIMIT_DEFAULTS`** holds the shipped policy — same numbers, same namespaces, no behaviour change out of the box. **`resolveRateLimitRules({ env, onWarn })`** layers overrides on top and is pure, so it can be tested against a fabricated environment without touching `process.env`.

**Variable names are derived from the rule name** rather than written out by hand:

```ts
envSuffixForRule("authForgotPassword")  // "AUTH_FORGOT_PASSWORD"
// -> RATE_LIMIT_AUTH_FORGOT_PASSWORD_LIMIT
//    RATE_LIMIT_AUTH_FORGOT_PASSWORD_WINDOW_SECONDS
```

That mapping was chosen to reproduce the five names already published in `.env.example`, so **existing deployments keep the variables they have** — there's a test pinning each one. It also means adding a rule can't leave it without an override, and renaming one can't leave a stale key behind.

**All eleven rules are configurable now**, not just the five auth ones. That includes `chat`, which fronts a per-request billed Gemini call and previously had no way to be tightened without a redeploy.

**Namespaces are deliberately not overridable.** They are Redis key prefixes — letting an env var move one would silently reset live counters or collide two unrelated rules into the same bucket.

**Bad values are loud.** An unset or blank variable is normal and silent. A *set but unusable* one (`abc`, `0`, `-5`, `Infinity`) warns and falls back, because silently serving the default is exactly how you end up believing a limit is tighter than it is. A fractional value rounds down and says so.

## Backwards compatibility

`RATE_LIMIT_CONFIG` still exists and still has its `{ auth: { signin: { limit, windowSeconds } } }` shape, so `signin.test.ts` and any other importer keeps compiling. It is now *derived* from the resolved rules and marked `@deprecated`, so the two tables can no longer drift.

## A note on the existing tests

`rate-limit-config.test.ts` asserted `signin.limit === 10` while reading `process.env`. Now that the value is genuinely env-driven, that assertion would fail on any machine with a tuned limit — and it was previously passing while asserting numbers nothing enforced. Default assertions now resolve against an explicit empty environment; a separate test asserts the exported view matches what the routes enforce.

Same treatment for `rate-limit-rules.test.ts`'s "preserves the limits the auth and chat routes already used" — it now reads `RATE_LIMIT_DEFAULTS`, since the guarantee is about the shipped defaults, not about a given deploy's config. All eight original tests are kept.

## Testing

```
Test Files  63 passed (63)
     Tests  443 passed (443)     # was 423
```

`npx tsc --noEmit` clean. 20 new tests covering name derivation, per-rule overrides, the untouched-namespace rule, every invalid-input path, and the config view tracking an override.

Verified manually that `RATE_LIMIT_AUTH_SIGNIN_LIMIT=2` now produces a 429 on the third sign-in attempt.


santichausis (approved):
Verified the load-bearing claim by hand: \`envSuffixForRule("authForgotPassword")\` does produce \`AUTH_FORGOT_PASSWORD\`, matching the five variable names already in \`.env.example\` — so this doesn't silently rename anyone's existing override. The deprecated \`RATE_LIMIT_CONFIG\` wrapper is now genuinely derived from the same resolved table rather than a second copy, which was the actual bug. Good call keeping namespaces non-overridable.

singh-odyssey (approved):

PR #439, opened 2026-08-28
security(pusher): fail closed when Pusher is unconfigured, and throttle the auth route

### LINKED ISSUE #434: [Security] POST /api/pusher/auth returns a fabricated auth signature when Pusher is unconfigured, and is unthrottled
## Summary

When `pusherServer` is null — which happens whenever `PUSHER_APP_ID` / `PUSHER_SECRET` / `PUSHER_KEY` are missing or malformed — `POST /api/pusher/auth` does not fail. It returns `200 OK` with a fabricated signature:

```ts
if (!pusherServer) {
  // Return mock auth response for development/testing if Pusher credentials are not set
  return NextResponse.json({
    auth: `mock-auth-signature-for-${userId}`,
  });
}
```

There is no environment guard on that branch. A production deployment that loses its Pusher secret serves the mock path to real users.

## Impact

1. **Silent failure in production.** Every realtime subscribe returns 200 with a signature Pusher will reject. The client sees a successful auth call and a channel that never delivers, so messages and connection requests just stop arriving with nothing in the logs and nothing in the error tracker. The failure mode is indistinguishable from "the app is quiet".
2. **The response body echoes the caller's own user id** back into a string that reads like a credential. Anything that logs or caches auth responses now stores a token-shaped value keyed to a real account.
3. **The endpoint is unthrottled.** It is the only authenticated route that runs a database query per call with no `enforceRateLimit`. `private-chat-<id>` runs a `conversation.findFirst` on every request, so a loop over guessed conversation ids is a free database scan and a membership oracle.
4. **No bounds on `channel_name`.** The schema is `z.string()` with no `max()`. A multi-megabyte channel name is accepted, hashed and sent to `authorizeChannel`.
5. **Sloppy channel parsing.** `channel_name.replace("private-user-", "")` on `"private-user-"` yields `""`, and on `"private-user-private-user-x"` strips only the first occurrence. Neither is exploitable today because the equality check catches them, but the parsing does not express the intended grammar.
6. **Soft-deleted accounts keep their subscriptions.** `/api/user/account` sets `isDeleted`, but a still-valid session token authorises channels here — the rest of the API checks `isDeleted: false`, this route does not.

## Steps to reproduce

1. Unset `PUSHER_APP_ID` and restart.
2. Sign in, then `POST /api/pusher/auth` with `socket_id=123.456` and `channel_name=private-user-<your id>`.
3. Actual: `200 {"auth":"mock-auth-signature-for-<your id>"}`. Expected: a 503 that says realtime is unavailable.

## Proposed fix

- Fail closed. Keep the mock only behind an explicit opt-in (`NODE_ENV === "test"` or `PUSHER_ALLOW_MOCK_AUTH=true`), and return `503` with a clear message otherwise.
- Parse channel names with an anchored regex rather than `startsWith` + `replace`, and reject empty or malformed ids.
- Bound `socket_id` and `channel_name`, and validate the `socket_id` shape Pusher documents (`\d+\.\d+`).
- Add a `pusherAuth` rate-limit rule keyed on the session user.
- Reject soft-deleted accounts, matching every other authenticated route.
- Cover all of it with tests — there is currently no test file for this route.

MOHITKOURAV01:
Closes #434

## The problem

When `pusherServer` was null — any deployment missing or fumbling `PUSHER_APP_ID` / `PUSHER_KEY` / `PUSHER_SECRET` / `PUSHER_CLUSTER` — the route did not fail. It returned `200 OK`:

```ts
if (!pusherServer) {
  // Return mock auth response for development/testing if Pusher credentials are not set
  return NextResponse.json({
    auth: `mock-auth-signature-for-${userId}`,
  });
}
```

No environment guard. A production box that loses its Pusher secret serves that to real users, Pusher rejects the signature, and realtime silently dies — the client sees a successful auth call and a channel that never delivers. The failure is indistinguishable from "the app is quiet", which is the worst way for messaging to break. The body also wrote a real account id into a field shaped like a credential, so anything logging or caching auth responses retained it.

## Changes

**Fail closed.** The mock now requires `isMockAuthAllowed()` — `NODE_ENV` of `test` or `development`, or an explicit `PUSHER_ALLOW_MOCK_AUTH=true` for a staging box that wants it. Everything else gets `503` with a message the client can surface and retry against. Note this means Vercel preview deployments (which run `NODE_ENV=production`) also fail closed, which is the intent. The mock body no longer carries the user id.

**Real channel-name parsing.** `startsWith(prefix)` + `replace(prefix, "")` does not express the grammar:

| input | old parse | now |
|---|---|---|
| `private-user-` | `""` | rejected |
| `private-user-private-user-x` | `private-user-x` | rejected |

Neither was exploitable, because the `!==` check downstream caught both — but the safety lived in a coincidence rather than in the parse. Anchored patterns now: `/^private-user-([A-Za-z0-9_-]{1,128})$/` and the chat equivalent.

**Bounds.** `socket_id` and `channel_name` were `z.string()` with no ceiling, so a multi-megabyte channel name was accepted, hashed for a rate-limit key and passed to `authorizeChannel`. Now capped at Pusher's documented 164 characters, with `socket_id` required to match `\d+\.\d+`.

**Throttling.** New `pusherAuth` rule (60/min, keyed on the session user). This was the only authenticated route running a database query per call with no limit — `private-chat-<id>` runs a `conversation.findFirst` every time, so looping over guessed conversation ids was a free scan *and* a membership oracle. 60/min sits well above a client reconnecting and authorising a handful of channels, and well below what a scan needs.

**Soft-deleted accounts refused.** `/api/user/account` sets `isDeleted`, but the session token stays valid until it expires. Every other authenticated route filters on `isDeleted: false`; this one didn't, so a deleted user kept receiving realtime traffic.

## Note on file layout

`isMockAuthAllowed` lives in `src/lib/pusher.ts`, not the route. Next.js only permits handler exports from a route module — keeping it local failed the build with `Property 'isMockAuthAllowed' is incompatible with index signature`.

## Testing

This endpoint had no test file. 23 new tests: ownership on both channel families, the five rejected channel shapes, socket_id validation, the 503 and mock paths asserted separately (including that neither body leaks the user id), soft-delete refusal, throttling headers, and `isMockAuthAllowed` across production/dev/opt-in.

```
Test Files  64 passed (64)
     Tests  446 passed (446)
```

`npx tsc --noEmit` clean.

### Interaction with #438

This registers a new rule in `src/lib/rate-limit-rules.ts`, which #438 also restructures. I verified the combination rather than guessing: merging both in either order is **conflict-free**, and the integrated tree typechecks with the full suite green. `pusherAuth` lands inside #438's `RATE_LIMIT_DEFAULTS`, so it picks up `RATE_LIMIT_PUSHER_AUTH_LIMIT` / `_WINDOW_SECONDS` overrides for free.

One small follow-up if #438 lands first: that PR's `.env.example` section documents every rule, and this rule should get its two commented lines added there.


santichausis (approved):
Read the route in full. The anchored regex replacing \`startsWith\`+\`replace\` is the right fix even though the old check wasn't exploitable — good instinct to not leave correctness resting on a coincidence. Fail-closed behavior, the soft-delete check, and dropping the user id from the mock body all check out against the diff, not just the description.

singh-odyssey (approved):

PR #440, opened 2026-08-28
fix(report): bound the report payload, enforce a reason set, and stop duplicate filings

### LINKED ISSUE #435: [Bug] POST /api/user/report accepts unbounded free-text reports and unlimited duplicates against the same user
## Summary

`POST /api/user/report` is the moderation intake for a platform whose whole premise is verified, safe travel. Its validation is three lines and none of them bound anything:

```ts
const ReportSchema = z.object({
  reportedId: z.string().min(1, "Reported user ID required"),
  reason: z.string().min(1, "Reason for report required"),
  details: z.string().optional(),
});
```

`reason` is a `String` column and `details` is `@db.Text`. Both accept unlimited length, and the same reporter can file the same report against the same person forever.

## Problems

**1. Unbounded payload.** `reason` and `details` have no `max()`. A single request can write a multi-megabyte `details` blob. The rate limit is 5 per 10 minutes per user, which caps requests, not bytes — five requests still write as much as the body parser allows, and every one of them lands in the table a moderator has to page through.

**2. `reason` is free text where it should be an enum.** Every report arrives as an arbitrary string, so the moderation queue cannot be grouped, counted or filtered by category. Two reports of the same behaviour spelled `"harassment"` and `"Harassment "` are different values.

**3. No duplicate guard.** Nothing stops the same `(reporterId, reportedId)` pair being inserted repeatedly. Within the rate limit a reporter files 5 identical reports every 10 minutes — 720 a day against one person. There is no unique constraint and no application-level check, so a coordinated group can bury a legitimate user under volume and there is no signal distinguishing that from 720 genuine complaints.

**4. Existence oracle.** An unknown `reportedId` returns `404 "User to report not found"` while a real one returns `201`. That turns the endpoint into a probe for whether a given user id exists — the same class of leak that `/api/users` was hardened against when `email` was removed from its search columns.

**5. Soft-deleted targets accepted.** The lookup is `prisma.user.findUnique({ where: { id: reportedId } })` with no `isDeleted: false`. Reports accumulate against deleted accounts that no moderator can action.

**6. Blocking is ignored.** `reportedId` is never checked against the block table, and self-reports are caught but reports naming a user who has already been blocked still create rows.

## Steps to reproduce

```
POST /api/user/report
{"reportedId":"<real id>","reason":"x","details":"<5 MB of 'a'>"}
```
`201 Created`. Repeat 4 more times — all five succeed and all five rows persist.

## Proposed fix

- Replace `reason` with a closed enum (`HARASSMENT`, `SPAM`, `SCAM`, `IMPERSONATION`, `INAPPROPRIATE_CONTENT`, `SAFETY_CONCERN`, `OTHER`) and cap `details` at 1,000 characters.
- Reject a duplicate `(reporterId, reportedId)` report inside a cooldown window with `409` and a clear message, rather than silently stacking rows.
- Scope the target lookup to `isDeleted: false`.
- Return the same response for a missing target as for a successful submission, so the endpoint stops answering "does this id exist".
- Add tests — this route has no test file today.

MOHITKOURAV01:
Closes #435

## The problem

This is the moderation intake for a platform whose premise is verified, safe travel. Its validation was:

```ts
const ReportSchema = z.object({
  reportedId: z.string().min(1, "Reported user ID required"),
  reason: z.string().min(1, "Reason for report required"),
  details: z.string().optional(),
});
```

`reason` is a `String` column, `details` is `@db.Text`, and neither had a ceiling.

### `details` is capped at 1,000 characters

No `max()` meant one request could write a multi-megabyte blob. The rate limit (5 per 10 minutes) caps *requests*, not bytes — five requests still store as much as the body parser allows. 1,000 characters is several paragraphs, comfortably more than the three-row textarea invites. The form now shows a live `n/1000` counter so the limit is visible before submit rather than arriving as a 400.

### `reason` becomes a closed set

Seven codes: `HARASSMENT`, `SPAM`, `SCAM`, `IMPERSONATION`, `INAPPROPRIATE_CONTENT`, `SAFETY_CONCERN`, `OTHER`.

The root cause is in the UI — the `<select>` posted its *display strings* straight to the API:

```tsx
<option value="Inappropriate behavior">Inappropriate behavior</option>
```

So the column filled with free text, and rewording an option silently created a new category. The modal now posts codes and renders labels, and the two are separated so the wording can change without rewriting history.

**Old clients keep working.** `normalizeReportReason` accepts the five legacy display strings (case-insensitive, whitespace-tolerant) as well as codes, so a browser holding the previous bundle submits successfully through a deploy instead of getting 400s.

**No migration.** `reason` stays a `String`, so existing rows are untouched and there is no backfill to coordinate. Promoting it to a Prisma enum would be a reasonable follow-up once historical values are normalised.

### Duplicates are refused

Nothing stopped the same `(reporterId, reportedId)` pair being inserted repeatedly — within the rate limit, 720 identical reports a day against one person. That buries a legitimate account in volume and leaves no signal distinguishing it from 720 genuine complaints. A second report against the same user inside 24 hours now returns `409` with the original report id.

### The existence oracle is closed

An unknown `reportedId` returned `404 "User to report not found"`; a real one returned `201`. That is a probe for whether any given user id exists — the same class of leak `/api/users` was hardened against when `email` came out of its search columns. Both cases now return the same `201 { success: true }`; only the real one writes a row.

### Soft-deleted targets excluded

The lookup moves to `findFirst({ where: { id, isDeleted: false } })`. Reports against deleted accounts can't be actioned, so they were pure queue noise.

## A note on an existing test

`moderation.test.ts` asserted `reason: "Spam"` passing through verbatim and a `findUnique` lookup. Both are precisely the behaviour being changed, so that test is updated to the new contract rather than worked around, and a case covering rejection of free-text reasons is added beside it.

## Testing

33 new tests in a dedicated file (the route had none): every code accepted, every legacy string mapped, four rejection shapes, cap boundaries including exactly-at-limit and a 5 MB body, the duplicate window and its scoping, the oracle behaviour asserted on both status *and* body shape, soft-delete exclusion, and throttling.

```
Test Files  64 passed (64)
     Tests  457 passed (457)
```

`npx tsc --noEmit` clean.


santichausis (approved):
Read \`report-reasons.ts\` and the route in full. The existence-oracle fix is real (unknown and real-but-untouched targets both return the same 201), and the legacy alias table matches what's actually live in \`ReportUserModal.tsx\` right now, not an older version of the strings — so in-flight clients won't get 400s through the deploy. Duplicate-window check is scoped to the reporter, not a global lock, which is correct.

PR #441, opened 2026-08-28
security(auth): stop the OTP routes confirming which addresses have accounts

### LINKED ISSUE #436: [Security] OTP endpoints leak account existence and let an attacker flood a chosen inbox
## Summary

`/api/auth/resend-otp` and `/api/auth/verify-otp` both answer, to an unauthenticated caller, whether an email address has an account and what state that account is in. `resend-otp` additionally sends a real email on every accepted call, so the same oracle doubles as a way to flood a chosen inbox.

## The oracle

Both routes branch on the account before doing any work:

```ts
const user = await prisma.user.findUnique({ where: { email } });
if (!user) {
  return ... { error: "User not found" }, { status: 404 }
}
if (user.emailVerified) {
  return ... { error: "Email already verified" }, { status: 400 }
}
```

Three distinguishable answers for one input:

| Response | What the caller learns |
|---|---|
| `404 User not found` | no account |
| `400 Email already verified` | account exists, verified |
| `200 New OTP sent` | account exists, **not yet verified** |

`/api/auth/forgot-password` already gets this right and returns the same body regardless. These two were never brought in line. The third bucket is the damaging one — it identifies half-finished signups, which are exactly the accounts whose email is confirmed to be reachable and whose owner is already expecting an OTP mail.

## The flooding

`resendOtp` is limited to 3 per 10 minutes, and `getRateLimitIdentifier(request, email)` keys the bucket on `email|ip`. The IP is in the key, so the limit is per address *per IP*. Rotating source addresses multiplies it directly: 3 requests per IP per window, and each one calls `sendOTPEmail`. There is no cooldown stored against the account itself, so nothing in the database caps how often a single mailbox is written to.

Each accepted call also overwrites `otp` and `otpExpires`. A victim mid-signup has the code in their inbox invalidated by every attacker request, which blocks the account from completing verification for as long as the attacker keeps going.

## Related weakness in verify-otp

`verify-otp` has no per-account attempt counter. The rate limit is keyed on `email|ip`, so 10 guesses per 10 minutes per IP against a 6-digit code, with no lockout and no invalidation of the OTP after repeated failures. The code is not cleared or rotated on failure, so a fixed target stays guessable for the full 10-minute lifetime across as many IPs as the attacker has.

## Steps to reproduce

```
POST /api/auth/resend-otp {"email":"nobody@example.com"}   -> 404
POST /api/auth/resend-otp {"email":"verified@example.com"} -> 400 "Email already verified"
POST /api/auth/resend-otp {"email":"pending@example.com"}  -> 200 + an email is sent
```

## Proposed fix

- Return one generic `200 { ok: true }` from `resend-otp` for every input, matching `forgot-password`. Do the send only for a real unverified account.
- Collapse `verify-otp`'s `404 User not found` into the same `400 Invalid or expired OTP` the bad-code path returns.
- Add a database-backed resend cooldown on `otpExpires` so a mailbox cannot be written to more than once per interval regardless of source IP.
- Track failed OTP attempts per account and invalidate the code once a threshold is crossed, so guessing costs a new email rather than another try.
- Tests for each response path.

MOHITKOURAV01:
Closes #436

## The oracle

Both routes branched on the account before doing any work, giving three distinguishable answers for one input:

| Response | What the caller learns |
|---|---|
| `404 User not found` | no account |
| `400 Email already verified` | account exists, verified |
| `200 New OTP sent` | account exists, **not yet verified** |

The third bucket is the damaging one. It identifies half-finished signups — accounts whose address is confirmed reachable and whose owner is *already expecting* an OTP mail. `/api/auth/forgot-password` has always returned a generic response regardless of outcome; these two were never brought in line.

**`resend-otp`** now returns one generic `200` for every input. The send happens only for a real unverified account, and **a delivery failure does not change the response** — otherwise an SMTP error against a real address re-opens the oracle from the other side.

**`verify-otp`** collapses `"User not found"`, `"No OTP found"` and `"OTP has expired"` into a single message.

`"Email already verified"` goes with them, which is worth calling out because it wasn't in the original issue plan. That branch runs *before* the code is compared, so keeping its own message would answer "does this address have a verified account?" to anyone submitting six random digits — the same oracle, one branch further down. A verified account has had its `otp` cleared, so a legitimate double-submit would land on the `!user.otp` check and get this identical message anyway. Nothing is lost.

## The flooding

`authResendOtp` is 3 per 10 minutes keyed on `email|ip`. The IP is *in the key*, so rotating source addresses multiplies the allowance directly, and `checkRateLimit` deliberately fails open when Redis is unreachable. Nothing in the database capped how often a single mailbox was written to.

Worse, every accepted call overwrote `otp` and `otpExpires`. A victim mid-signup had the code in their inbox invalidated by each attacker request — so the account could be blocked from ever completing verification for as long as the attacker kept going.

`canResendOTP` adds a 60-second cooldown derived from **stored state**:

```ts
export function otpIssuedAt(otpExpires: Date | null | undefined): Date | null {
  if (!otpExpires) return null;
  return new Date(otpExpires.getTime() - OTP_TTL_MS);
}
```

There is no `otpIssuedAt` column, but the issue time is recoverable from the expiry because every code is written with the same TTL — so this needs no migration. Because it reads the database rather than Redis, it holds regardless of source IP *and* when the limiter has failed open (there's a test for exactly that). Inside the cooldown the stored code is left untouched, so an attacker can no longer invalidate a victim's live code.

A future issue time (clock skew, a hand-edited row) is treated as "just issued" rather than trusted to unlock a resend.

## The guessing

`authVerifyOtp` is likewise keyed on `email|ip` — 10 attempts per address per window against a 6-digit code, unbounded in aggregate. Added `authVerifyOtpAccount` (20 per 15 min), keyed on the **email alone** via a new option:

```ts
await enforceRateLimit(req, "authVerifyOtpAccount", email, { scope: "subject" });
```

`RateLimitScope` defaults to `subject-and-ip`, so every existing caller is unchanged. The doc comment spells out when dropping the IP is correct: only where the cost of an attempt lands on somebody other than the caller.

## Client compatibility

`sign-up-form.tsx` reads `data.message` on success and `data.error` on failure. Both still present, no client change needed.

## Testing

22 new tests; neither route had any. The headline one asserts the three responses are now byte-identical rather than checking each in isolation — that's the actual property under test.

```
Test Files  64 passed (64)
     Tests  445 passed (445)
```

`npx tsc --noEmit` clean.

### Interaction with #438 and #439

All three touch `src/lib/rate-limit-rules.ts`. I verified the combination rather than guessing: merging all three in either order is **conflict-free**, and the integrated tree typechecks with the full suite green. The `scope` parameter added here and #438's env-driven resolution compose cleanly — `enforceRateLimit` ends up with both — and `authVerifyOtpAccount` lands inside `RATE_LIMIT_DEFAULTS`, so it picks up env overrides for free.

One small follow-up if #438 lands first: that PR's `.env.example` section documents every rule, and `authVerifyOtpAccount` should get its two commented lines added there.


santichausis (approved):
This is the one I checked most carefully, since collapsing an oracle is easy to get almost right. Every failure branch in \`verify-otp\` — bad format, unknown user, already verified, no OTP, expired, wrong code — converges on the same \`invalid()\` response, and the reasoning for folding "already verified" in with the rest (it runs before the code comparison, so it's the same leak one branch later) is correct. \`canResendOTP\` reading the cooldown from \`otpExpires - OTP_TTL_MS\` instead of needing a new column is a clean way to avoid a migration. The \`subject\`-scoped rate limit is the right tool for an account-following ceiling, and defaults to the existing \`subject-and-ip\` behavior for every other caller.

PR #442, opened 2026-08-28
fix(notifications): route both mark-as-read endpoints through the ownership-scoped helper

### LINKED ISSUE #437: [Bug] Duplicate notification mark-as-read routes bypass the ownership-scoped helper and can mutate expired rows
## Summary

There are two routes that mark a notification as read, they are copies of each other, and neither uses the helper written for exactly this job. The bell component calls both.

- `PATCH /api/notifications/[id]`      — `src/app/api/notifications/[id]/route.ts`
- `PATCH /api/notifications/[id]/read` — `src/app/api/notifications/[id]/read/route.ts`

`src/components/notification-bell.tsx` hits the first on line 159 and the second on line 175.

## The dead helper

`src/lib/notifications.ts` exports:

```ts
export async function markNotificationAsRead(id: string, userId: string) {
  return prisma.notification.updateMany({
    where: { id, userId },
    data: { read: true },
  });
}
```

```
$ grep -rn "markNotificationAsRead" src
src/lib/notifications.ts:223:export async function markNotificationAsRead(
```

No callers. Both routes hand-roll the same operation instead — and hand-roll it worse.

## Why the hand-rolled version is worse

Both routes do a check followed by an unscoped write:

```ts
const notification = await prisma.notification.findFirst({
  where: { id: params.id, userId: session.user.id },   // ownership checked here
});
if (!notification) return 404;

await prisma.notification.update({
  where: { id: params.id },                            // ...but not here
  data: { read: true },
});
```

1. **Two round trips where one suffices.** `updateMany({ where: { id, userId } })` does the check and the write in a single scoped statement, which is precisely why the helper exists and why `deleteNotification` right beside it in the same file is written that way.
2. **The write is not ownership-scoped.** The `update` matches on `id` alone. It is correct today only because the `findFirst` ran first — the safety lives in the sequencing, not in the query. `deleteNotification` does not rely on sequencing; these two do. Any future edit that reorders or short-circuits the guard turns it into a cross-account write with no failing test to catch it.
3. **`RecordNotFound` becomes a 500.** If the row is deleted between the two statements — a real interleaving, since `DELETE /api/notifications` and the expiry sweeper both remove rows concurrently — Prisma throws `P2025` and the handler returns `500 Failed to update notification`. Marking an already-gone notification read should be a 404.
4. **Expired notifications are readable.** `GET /api/notifications` filters through `activeNotificationWhere()` so expired rows never render. Neither PATCH applies that filter, so an id captured before expiry can still be mutated after it. The read path and the write path disagree about which rows exist.

## Duplication

The two files are the same handler twice, one returning `{ ok: true, notification }` and the other `{ success: true }`. Two shapes for one action, and the client uses both, so a component reading `res.success` works on one route and silently no-ops on the other.

## Proposed fix

- Route both handlers through `markNotificationAsRead`, extended to take the `now` clock and apply `activeNotificationWhere()`.
- Return `404` on a zero `count` — no `findFirst`, no `P2025`, one statement.
- Keep `/[id]/read` as a thin re-export of the canonical handler so the existing client call keeps working, and align both response bodies on `{ ok: true, notification }` with `success` retained as an alias.
- Point `notification-bell.tsx` at one endpoint.
- Add tests for ownership, expiry, the concurrent-delete path and the response shape.

MOHITKOURAV01:
Closes #437

## The dead helper

`src/lib/notifications.ts` exports `markNotificationAsRead`. Nothing called it:

```
$ grep -rn "markNotificationAsRead" src
src/lib/notifications.ts:223:export async function markNotificationAsRead(
```

Two routes marked a notification read — `PATCH /api/notifications/[id]` and `PATCH /api/notifications/[id]/read` — and both hand-rolled the operation instead. The bell called one on line 159 and the other on line 175.

## Why the hand-rolled version was worse

```ts
const notification = await prisma.notification.findFirst({
  where: { id: params.id, userId: session.user.id },   // ownership checked here
});
if (!notification) return 404;

await prisma.notification.update({
  where: { id: params.id },                            // ...but not here
  data: { read: true },
});
```

**The write isn't ownership-scoped.** It's correct today *only because the `findFirst` ran first* — the safety lives in the sequencing, not the query. `deleteNotification`, ten lines above it in the same file, doesn't rely on sequencing. These did, and no test would have caught an edit that reordered them.

**A concurrent delete became a 500.** `DELETE /api/notifications` and the expiry sweeper both remove rows while this runs. Prisma throws `P2025` between the two statements and the handler reported `500 Failed to update notification`. Marking an already-gone notification read is a 404.

**The read and write paths disagreed about which rows exist.** `GET /api/notifications` filters through `activeNotificationWhere()` so expired rows never render — neither PATCH applied it, so an id captured before expiry stayed mutable after it.

**Two shapes for one action.** `{ ok: true, notification }` from one, `{ success: true }` from the other. A component reading `res.success` worked against one endpoint and silently no-opped against the other.

## The change

`markNotificationAsRead(id, userId, now)` now takes the clock, applies `activeNotificationWhere()`, and returns the row or `null`:

```ts
const result = await prisma.notification.updateMany({
  where: { id, userId, ...activeNotificationWhere(now) },
  data: { read: true },
});
if (result.count === 0) return null;
```

One scoped statement does the check and the write. A foreign id, a missing one and an expired one are indistinguishable to the caller, which is what we want.

To be straight about the query count: the *write* is one statement, but building the response body still needs a read of the updated row. That read is only reached on success and is already known to be the caller's row — the ownership decision is made entirely by the `updateMany`. On the 404 path there's no second query at all, which is asserted.

`/[id]` is the canonical handler. `/[id]/read` **re-exports** it rather than being deleted — the path is public API and any client still on it should keep working, and re-exporting means they cannot drift apart again. There's a test asserting the two exported handlers are literally the same function.

The response carries **both** `ok` and `success`, so clients reading either previous shape keep working. `notification-bell.tsx` now calls one endpoint.

## Testing

14 new tests across two files: ownership scoping asserted on the `updateMany` `where` clause (not just on the outcome), the expiry filter, the concurrent-delete path, the 404-without-second-query case, and the alias identity.

```
Test Files  65 passed (65)
     Tests  437 passed (437)
```

`npx tsc --noEmit` clean.


santichausis (approved):
Read \`markNotificationAsRead\` and both routes. The ownership check and the write are now one \`updateMany\` statement instead of a \`findFirst\`-then-\`update\` pair that was only correct because of sequencing — and \`activeNotificationWhere\` being applied there closes the read/write disagreement on expired rows. Re-exporting the alias route rather than deleting it is the right call for a path that's public API.

singh-odyssey (approved):

PR #404, opened 2026-08-16
feat(meetup-plans): implement PATCH and DELETE handlers for meetup plans

Pratyush-Panda-2006:
Fixes #399

### Summary
Implemented \PATCH\ and \DELETE\ route handlers in \src/app/api/meetup-plans/route.ts\. Users can now modify plan details (title, location, date/time, coordinates, notes, route) and delete existing plans with proper authorization checks.

### Verification
Added comprehensive unit tests in \src/app/api/meetup-plans/__tests__/meetup-plans-crud.test.ts\ verifying auth, membership checks, field updates, and deletion.

PR #444, opened 2026-09-21
test: fix two stale tests that turned main's CI red

santichausis:
`CI` on `main` has been failing since the #403 / #404 / #406 merges (8 tests across 2 files, `build-and-lint` → Test step).

## Cause

**1. `src/app/api/user/block/__tests__/route.test.ts` (1 test)** — regression from #406.
#406 intentionally removed `status: "PENDING"` from the `connectionRequest.deleteMany` where-clause so blocking also clears an ACCEPTED connection. The pre-existing test still asserted the old filter. It wasn't in #406's diff, and that branch never had CI run (`action_required` on the fork PR), so it only surfaced after merge.

**2. `src/app/api/meetup-plans/__tests__/route.test.ts` (7 tests)** — time bomb, not caused by #404.
`VALID_BODY.meetupTime` was hardcoded to `2026-09-01T09:00:00Z`. `POST` rejects a meetup more than a day in the past (`MAX_MEETUP_BACKDATE_MS`), so once the calendar passed 2026-09-02 every valid-body POST test returned `400` instead of `201`/`403`/`409`. `main` would have gone red on any PR after that date, with or without #404.

## Fix (tests only, no production code)

- block test: assert the new behaviour — `deleteMany` with **no** `status` filter (exact-match assertion, so re-adding a status filter fails the test) — and rename it accordingly.
- meetup test: `meetupTime` is now `Date.now() + 24h`; the `create` assertion compares against `VALID_BODY.meetupTime` instead of a second hardcoded date.

## Verification

Reproduced the same 8 failures locally on `origin/main` before changing anything, then:

```
$ npx vitest run
Test Files  74 passed (74)
     Tests  613 passed (613)

$ npx tsc --noEmit     # clean
$ npm run lint         # no warnings or errors
```


santichausis:
@singh-odyssey could you review and merge this when you get a chance? `main` has been red since the #404 / #406 merges on 09/14 (last failed run: 09/15 after #443), and this puts it back to green.

It's tests only, no production code:
- `user/block` test still asserted the old `status: "PENDING"` filter that #406 removed on purpose.
- `meetup-plans` test had a hardcoded `meetupTime` that started failing once it fell more than a day in the past.

CI on this PR passes (`build-and-lint`), and none of the open PRs touch these files, so there's no conflict risk. I can't approve my own PR, which is why I'm asking.

singh-odyssey (approved):

